PR #27, opened 2014-11-24
Changed allowed branches from fixed select to free form textbox.

FrantaM:
Current approach puts unnecessary conditions to branch filtering:

* Access to gitlab must be set in system config.
* Configured gitlab user must see the project.
* Selected branch must be available in gitlab when configuring project in jenkins.

this fails in basic scenario when you *don't* have a separate gitlab user and use deploy keys to checkout sources in jenkins.


FrantaM:
>Current approach puts unnecessary conditions to branch filtering:
>
>* Access to gitlab must be set in system config.
>* Configured gitlab user must see the project.

This is correct, and is now being used by more features such as linking to build pages from GitLab. 


> Selected branch must be available in gitlab when configuring project in jenkins.

It is still possible to add a branch to the filter later by updating the project configuration in jenkins. This you will have to do at some point any way you approach filtering (although regexes might make life easier with this).


> this fails in basic scenario when you don't have a separate gitlab user and use deploy keys to checkout sources in jenkins.

Using deploy keys or a gitlab user's access token are similar approaches, is there a reason for chosing one over the other?

If I understand your approach correctly, it is a text box filled with branch names? Other plugins opted for a regex solution, which is similar and I had already considered that in the beginning. The reason I went for a list of branches is that it's easier to configure. Any project with more than a couple of branches would make filtering with a text box quite inconvenient: long strings, complicated regexes, trying to remember branch names etc etc


so while I do see your point about "unnecessary" requirements on filtering, I'd like to keep the approach as it is now; if configuring access to gitlab is not possible, then only the basic functionality of building will be available. It seems to me that your scenario can be served by git-level filtering (something similar to https://wiki.jenkins-ci.org/display/JENKINS/Git+Parameter+Plugin perhaps?).

FrantaM:
> This is correct, and is now being used by more features such as linking to build pages from GitLab.

You see I think that access to gitlab should be _supported_ not _required_ unless necessary. Both jenkins and gitlab may not be maintained by the same person. This could extend the time to setup a simple job from several minutes to days. You should also consider that the gitlab admin may not we willing (or even able) to create a separate user at all.
 
> It is still possible to add a branch to the filter later by updating the project configuration in jenkins.

This means you’re not able to setup a job properly before you push that branch. Also when you open the configure page and a already-configured branch is not present in gitlab (or the connection is lost or whatever) it won’t be present in the selectbox will just slip out. Even if you notice this you won’t be able to put it back.
 
> Using deploy keys or a gitlab user's access token are similar approaches, is there a reason for chosing one over the other?

If there weren’t any there wouldn’t be both.
 
You understand that correctly, it is a textbox filled with branch names (with autocomplete). Regex or ant-pattern solution would even be better. Any project with more than a couple of branches would be much easier to setup with this. Trust me, I’ve been there. If you have a large project you need flexibility and don’t mind complexity that much.

bassrock:
@FrantaM can you rebase this based on the latest pulls. Can you also provide examples for your regex solution in the jelly help question mark? Will it be if in the regex they are built? Also can you default the reggae to include all branches? I like this approach better, but also want feedback from the community already using this.

xaniasd:
how about asking on Gitter?

bassrock:
was just about to

massimone88:
Hi! I'm interested about this feature. How is the road map for this pull request?

bassrock:
@massimone88 we need it rebased on the latest so it can be merged

bassrock:
@FrantaM looks like i didn't notice your rebase and didn't get notified, looks like changes were made that requires another rebase.

massimone88:
news?

PR #36, opened 2014-12-21
Locate build by data from git plugin

kasper-f:
Changed getBuildBySHA1 and getBuildByBranch to use the git plugins MergerRecord and BuildData to determine if the build is a merge what SHA1/branch the build cover.

This way we can ensure that we locate the correct build even in the cases where the ParameterActions is not present in the build.

The getBuildByBranch will only return builds that is not build from a merge.

[MergeRecord] (https://github.com/jenkinsci/git-plugin/blob/master/src/main/java/hudson/plugins/git/util/MergeRecord.java)
[BuildData] (https://github.com/jenkinsci/git-plugin/blob/master/src/main/java/hudson/plugins/git/util/BuildData.java)
[Build](https://github.com/jenkinsci/git-plugin/blob/master/src/main/java/hudson/plugins/git/util/Build.java)

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/GitLabWebHook.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/GitLabWebHook.java
@@ -13,6 +13,7 @@ import hudson.plugins.git.Branch;
 import hudson.plugins.git.GitSCM;
 import hudson.plugins.git.util.Build;
 import hudson.plugins.git.util.BuildData;
+import hudson.plugins.git.util.MergeRecord;
 import hudson.scm.SCM;
 import hudson.security.ACL;
 import hudson.util.HttpResponses;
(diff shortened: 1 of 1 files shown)


kasper-f:
Hi @xaniasd as agreed in pull request #32 I have opened this a new pull request against the develop branch. 
If there is any problem in the implementation please let me know and I will try fixing them?

xaniasd:
@kasper-f I'll test it for this weekend's release!

bassrock:
@kasper-f will test this tonight. Need to get a test environment up and running for this.

kasper-f:
Sounds good @bassrock . In case of any troubles just let me know. we have used this changes for the last month without noticing any  problems

bassrock:
Ok cool and this is able to locate merge request builds properly still too?

kasper-f:
@bassrock I have tested this with a small project containing 3 different merge request where I during the last moth have pushed to all three branches and haven't found any cases where the wrong build was found. So in my opinion this changes locate the merge request builds correctly as well.

bassrock:
So I was trying to test this but when running mvn hpi:run it seems like the dependencies are not getting copied and causing class path errors with the Gitlab API.  This warning also pops up 

```
[WARNING] The POM for org.jenkins-ci.tools:maven-hpi-plugin:jar:1.106 is missing, no dependency information available
[WARNING] Failed to build parent project for org.jenkins-ci.plugins:gitlab-plugin:jar:1.1.10-SNAPSHOT
```

Any ideas?

kasper-f:
@bassrock. I think it is due to a problematic merge in the develop branch [mentioned here] (https://github.com/DABSquared/gitlab-plugin/pull/36#issuecomment-67982971) where the <packaging>hpi</packaging> have diapered from the pom.xml file. 
I will try to look at it when I get back in front of my computer.

Issue #37, opened 2015-01-12
status.json raises 500 error


eltados:
When viewing a merge request on Gitlab, Gitlab makes a request to Jenkins to get the status of the commit, however this raises a NullPointerException.

The request made by Gitlab looks like this:

``https://JENKINS_URL/project/PROJECT_NAME/commits/COMMIT_SHA/status.json?token=TOKEN``

And the NullPointerException traceback is:
``
Caused by: java.lang.NullPointerException
	at com.dabsquared.gitlabjenkins.GitLabWebHook.getBuildBySHA1(GitLabWebHook.java:440)
	at com.dabsquared.gitlabjenkins.GitLabWebHook.generateStatusJSON(GitLabWebHook.java:183)
	at com.dabsquared.gitlabjenkins.GitLabWebHook.getDynamic(GitLabWebHook.java:137)
	at sun.reflect.GeneratedMethodAccessor238.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.kohsuke.stapler.Function$InstanceFunction.invoke(Function.java:298)
	at org.kohsuke.stapler.Function.bindAndInvoke(Function.java:161)
	at org.kohsuke.stapler.MetaClass$12.dispatch(MetaClass.java:389)
	at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:746)

``

k-kawa:
I got the same error when the Jenkins test job was started by another Jenkins job which starts with Gitlab webhook.
It seems ok for me if the job kicked by Gitlab webhook and the job set as a Gitlab Ci are the same

pmalek:
We're experiencing the same issue but the scenario is a little bit different:
I push a new branch to gitlab, create a merge request and then everything works fine - I get the CI build status on Gitlab merge request page. Then I push another commit to the same branch on Gitlab, build triggers successfully on Jenkins but the status fails, meaning that I have this error:

```
Cannot connect to the CI server. Please check your settings and try again.
```

![image](https://cloud.githubusercontent.com/assets/739996/5855888/6ded3f9a-a23d-11e4-92e1-2c91ec827c1f.png)


[...]

PR #38, opened 2015-01-12
Update GitLabWebHook.java

eltados:
Fix the null pointer exception

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/GitLabWebHook.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/GitLabWebHook.java
@@ -439,7 +439,11 @@ public class GitLabWebHook implements UnprotectedRootAction {
             ParametersAction params = build.getAction(ParametersAction.class);
             StringParameterValue sourceBranch = (StringParameterValue) params.getParameter("gitlabSourceBranch");
             StringParameterValue targetBranch = (StringParameterValue) params.getParameter("gitlabTargetBranch");
-            boolean isMergeRequestBuild = !sourceBranch.value.equals(targetBranch.value);
+            boolean isMergeRequestBuild = ! ( 
+            		sourceBranch!=null  && sourceBranch.value != null &&
+            		targetBranch != null && targetBranch.value != null &&
+            		sourceBranch.value.equals(targetBranch.value)
+            );
             
             if (!triggeredByMergeRequest) {
 				if (isMergeRequestBuild)


eltados:
Resolves issue #37  ( I think )

kasper-f:
This Pull Request will not be needed in case Pull Request #36 is accepted as we then will use the data from the git plugin to determine what branch  sha1 numbers the build covers instead of the build parameters.

kasper-f:
@pmalek Yes, since I have made this changes I have not experienced any problems and as so the #37 is solved for me. 

Please note:  The entire content of the getBuildBySHA1 function is changed in my commit and that the pull request is __not__ yet accepted.

bassrock:
@eltados and @kasper-f so do we want to use this not or #36?

kasper-f:
@bassrock I think the changes in #36 is preferable as it uses the data from the git-plugin to determine the correct build instead of relying on the custom parameters (gitlabSourceBranch and gitlabTargetBranch).

On the long field I hopefully see the need for custom parameters as unnecessary when we find a way to start a merge build from code (without the parameters). I have tried figuring out the different actions used by the git-plugin to start the merge build but haven't succeeded yet.

bassrock:
@kasper-f interesting. Sorry for the questions been a while since I have actively maintained this. Does this mean that gitlabSourceBranch and targetBranch are no longer required, even for merge request builds?

kasper-f:
As fare as I can tell the only usage of this parameters (apart from the getBuildBySHA1 method where I have removed it) is in the onPost triggers used to start a build of the job.

As I found it a bit annoying that I had to setup this [parameters](https://github.com/DABSquared/gitlab-plugin#parameterized-builds) and use them in my [jobs git configuration](https://github.com/DABSquared/gitlab-plugin#using-it-with-a-job) my goal is to find out if it is possible to call the git-plugin actions directly instead of a ParametersAction and this way avoid the troubling configuration for on each new project I would like to build. Then we could reduce the configuration to be a checkbox in the trigger to select if we would like to build merge requests or not instead of setting up additional repository and branch specifier from the parameters.

bassrock:
Nice I like that thought and that is the way I wanted the original plugin to work but could not figure out how. I'll leave them in the readme for now then until they get eliminated fully.

bassrock:
This should not be needed do to #36

eltados:
Sorry for the delay , I agree this change request is no longer needed

PR #43, opened 2015-01-21
Set build description to build cause

### LINKED ISSUE #39: Set Job Description to the Build Cause
The plugin should provide an option to set the Build Description to the build cause that triggered it.  For Merge Request Builds, this would be the message that says:

GitLab Merge Request &lt;mergeRequestNumber&gt; : &lt;sourceBranch&gt; =&gt; &lt;destBranch&gt;

This allows a user to quickly scan the list of recent builds to see which merge requests are good and which have issues.

For a comparison, see the [Gitlab Merge Request Builder](https://wiki.jenkins-ci.org/display/JENKINS/Gitlab+Merge+Request+Builder+Plugin) plugin for an example of how it sets the build description.

kasper-f:
This is a candidate to fix #39 

Have added a run listener that will call the trigger functions when the build starts (and when it is completed (not currently used but kept the function in order to ease further functionality))

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/GitLabMergeCause.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/GitLabMergeCause.java
@@ -32,26 +32,4 @@ public class GitLabMergeCause extends SCMTrigger.SCMTriggerCause {
                 " => " + this.mergeRequest.getObjectAttribute().getTargetBranch();
     }
 
-
-    public String getHtmlDescription() {
-        GitLab git = new GitLab();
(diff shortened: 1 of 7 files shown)


kasper-f:
I have Considered updating the GitLabPushCause to include the last commit in the build ?

I also considered to introduce a new function in both the cause classes to get the short description as html. This way we could add a link to gitlab to the specified merge request or the specific commit.

kasper-f:
A screenshot to show how it looks like in jenkins project view:
jenkins_build_description

PR #45, opened 2015-01-22
Fixed #28 "Build status icon toggling" 

### LINKED ISSUE #28: Build status icon toggling
The build icon from the plugin seems to toggle every refresh

The first time it looks like this:
image

The second time it looks correct:
image

kasper-f:
by adding a flushBuffer of the request after the image is served.
This is a candidate to fix #28


kasper-f:
I was not able to get the cache disabled using the StamplerRequest serveFile. 

Instead I use the hudson.IOUtils to copy the file resource to the output stream. 

This implementation is inspired by the [embeddable-build-status-plugin](https://github.com/jenkinsci/embeddable-build-status-plugin)  [StatusImage](https://github.com/jenkinsci/embeddable-build-status-plugin/blob/master/src/main/java/org/jenkinsci/plugins/badge/StatusImage.java)

PR #68, opened 2015-04-09
build when tags are pushed

mrwanny:
I would like to allow the plugin to support tags
i.e. when an annotated tag is push it should trigger a build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,15 @@ Triggers from push events may be filtered based on the branch name, i.e. the bui
 
 This functionality requires accessing the Gitlab server (see [above](#configuring-access-to-gitlab)) and for the time being also a git repository url already saved in the project configuration. In other words, when creating a new project, the configuration needs to be saved *once* before being able to select the allowed branches. For existing projects, all branches are allowed to push by default.
 
+Build Tags
+================
+
+In order to build when a new tag is pushed:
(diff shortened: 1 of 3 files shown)


bassrock:
@mrwanny Please add a checkbox so users can disable this if they do not want it.

mrwanny:
I thought about that but I am not sure we need an additional checkbox.
Branches to build -> Branch Specifier (blank for 'any') is already controlling with ref need to be build.

bassrock:
Yea, but for instance my setup I have it building all branches, but I do not need it building a tag.

mrwanny:
I forgot, you also need to add +refs/tags/*:refs/remotes/origin/tags/* into the Advance tab :-)

bassrock:
?

mrwanny:
As far as I know, by default the git plugin does not fetch the tags ref, it only fetches the remote branches ref, see below 
http://erics-notes.blogspot.com/2013/05/jenkins-build-latest-git-tag.html

koglinjg:
+1, I believe what @mrwanny is saying about it not fetching, or building, on tag pushes without the refspec change is correct.

bassrock:
ok. Will be happy to merge, can we add some documentation about it first since its a bit confusing unlike some of the others features, in my opinion.

mrwanny:
I did add a small section into the REAME.md, please let me know if it is sufficient. Thanks

PR #70, opened 2015-04-10
Adding parameters to merge request builds

koglinjg:
gitlabMergeRequestID and gitlabTargetProjectID will now show up as parameters when a job is kicked off before of the merge request trigger. This allows interaction through the gitlab API within the job. To lookup the merge request or perform any other actions see the gitlab API doc: https://github.com/gitlabhq/gitlabhq/blob/master/doc/api/merge_requests.md#get-single-mr
Conflicts:
    src/main/java/com/dabsquared/gitlabjenkins/GitLabPushTrigger.java

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/GitLabPushTrigger.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/GitLabPushTrigger.java
@@ -194,6 +194,8 @@ public class GitLabPushTrigger extends Trigger<AbstractProject<?, ?>> {
                     LOGGER.log(Level.INFO, "Trying to get name and URL for job: {0} using project {1}", new String[]{job.getName(), getDesc().project.getName()});
                     String sourceRepoName = getDesc().getSourceRepoNameDefault(job);
                     String sourceRepoURL = getDesc().getSourceRepoURLDefault(job).toString();
+                    String mergeRequestID = req.getObjectAttribute().getId().toString();
+                    String targetProjectID = req.getObjectAttribute().getTargetProjectId().toString();
                     
                     if (!getDescriptor().getGitlabHostUrl().isEmpty()) {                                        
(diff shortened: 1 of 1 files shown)


bassrock:
Not quite sure I understand why this is needed. Can you explain a bit more.

koglinjg:
With these parameters, you can look up the merge request and perform actions like:

custom messages (output from spec / lint etc)
```POST /projects/:id/merge_request/:merge_request_id/comments```

Get and parse comments
```GET /projects/:id/merge_request/:merge_request_id/comments```

Get MR attributes (votes, author, assignee)
```GET /projects/:id/merge_request/:merge_request_id```

Accept the MR
```PUT /projects/:id/merge_request/:merge_request_id/merge```
If you are using a deploy key for the jenkins job, this would allow you to accept the MR using the API as a different user, based on the auth token passed with the request.

Eventually when they add labels to their API you can also apply labels like "passing" and "failing" specs.


See more about the API here:
https://github.com/gitlabhq/gitlabhq/blob/master/doc/api/README.md
https://github.com/gitlabhq/gitlabhq/blob/master/doc/api/merge_requests.md

koglinjg:
@bassrock, Any update on this?

bassrock:
@koglinjg Happy to merge. Since this adds parameters is it required to add to the setup of a job process? If so can you update the realm to reflect the new parameters and when they need to be added?

omehegan:
@koglinjg thanks! I would like to merge this. Can you resolve the existing conflicts (maybe rebase)?

omehegan:
@koglinjg ping?

coder-hugo:
These both parameters are already available via the GitLabMergeCause during the complete build cycle of a merge request build.
If you need them outside of the plugin we should add them as environment variables as they won't be persisted in the build result. IMHO this could be done with all the other build parameters as well.

PR #75, opened 2015-04-25
Extends jobs default parameter with gitlab plugin parameter

xhaggi:
Instead of only passing the gitlab plugin specific parameters to the
build this will load the jobs default parameters and extends it by the
gitlab plugin parameter. Now it's possible to use additional parameters
added to your job configuration.

It also fix the exception 'Unrecognized macro 'XXX' in '${XXX}'' which
is thrown if you've specified additional parameters for your job.


xhaggi:
@bassrock done .. don't know that ;)

Issue #56, opened 2015-02-11
Merge Request CI status not updating


ndcampbell:
On a Gitlab merge request the CI status stays at "CI build pending for 6f3ca78b. Build page". The Build page link does not go to the build status (just goes to a blank Jenkins page) and even when the build is successful this pending message is never changed.

bassrock:
Can you post any logs from Jenkins?

mildewonrice:
Here are some of the logs I have from the jenkins.log. This is when one of my builds kicked off from a Merge Request and failed. The CI Status in Gitlab never gets updated at all. Let me know if there is anything else I can provide. 

Feb 11, 2015 9:58:50 AM com.dabsquared.gitlabjenkins.GitLabWebHook getDynamic
WARNING: WebHook called.
Feb 11, 2015 9:58:54 AM com.dabsquared.gitlabjenkins.GitLabWebHook getDynamic
WARNING: WebHook called.
Feb 11, 2015 9:58:56 AM com.dabsquared.gitlabjenkins.GitLabWebHook getDynamic
WARNING: WebHook called.
Feb 11, 2015 9:59:21 AM com.dabsquared.gitlabjenkins.GitLabWebHook getDynamic
WARNING: WebHook called.
Feb 11, 2015 9:59:39 AM hudson.model.Run execute
INFO: Venus Git Build #26 main build action completed: FAILURE
Feb 11, 2015 9:59:40 AM org.terracotta.jenkins.plugins.postcompleted.PostCompletedRunListener onCompleted

ejhayes:
I'm running into the same issue as well:

`https://****/project/XXXX/commits/047c93e28773ad2122ff16bd0fae9574ae78995b`

Returns nothing

`https://****/project/XXXX/builds/047c93e28773ad2122ff16bd0fae9574ae78995b/status.json`

Returns:

`{
sha: "047c93e28773ad2122ff16bd0fae9574ae78995b",
status: "pending"
}`

bassrock:
@ejhayes  and @mildewonrice what version of Gitlab?

ejhayes:
I'm using GitLab 7.6.2

mildewonrice:
I am running 7.7.2-ee

Kurry:
This seems like more of a plugin issue rather than a gitlab issue:

1) The HTTP redirect from the SHA1 to build page never occurs 

2) After the fails or completes the status.json file does not update.

bhaku:
[...]

PR #76, opened 2015-04-27
Fix broken build links since GitLab 7.10

### LINKED ISSUE #74: Gitlab build links broken in Gitlab 7.10
I have just upgraded from Gitlab 7.8.4 to Gitlab 7.10.0 and now the build status on merge requests always shows "pending" and the links to "View build page" just return a blank page.

In the Jenkins log I see this when accessing the build page:
Apr 25, 2015 2:28:03 PM com.dabsquared.gitlabjenkins.GitLabWebHook getDynamic
WARNING: WebHook called.
Apr 25, 2015 2:28:03 PM com.dabsquared.gitlabjenkins.GitLabWebHook getDynamic
WARNING: Dynamic request mot met: First path: 'refs' late path: '02eb3e12f2517ed2a4e6b52ecf13dc78bf61da1d'

Looking at the URLs being built it seems the Gitlab CI integration has changed what it looks for.

In 7.8.4 the build page link would be of the form:
http://jenkins.host.com/project/Project_Name/commits/cda3a5c549392ce3e2141a4c8798ddc90b148089
In 7.10.0 the link is now:
http://jenkins.host.com/project/Project_Name/refs/sourceBranchName/commits/cda3a5c549392ce3e2141a4c8798ddc90b148089

This looks to be the change to gitlab which altered this functionality:
https://github.com/gitlabhq/gitlabhq/commit/5738afdb12162683f12b4fce8a4d6646ddca9c60

xhaggi:
This will fixes the broken build links since GitLab 7.10 (fixes #74 and #56)

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/GitLabWebHook.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/GitLabWebHook.java
@@ -108,6 +108,14 @@ public class GitLabWebHook implements UnprotectedRootAction {
         while (restOfPathParts.hasNext()) {
             paths.add(restOfPathParts.next());
         }
+        
+        // remove everything till we found 'commits'
+        for (Iterator<String> it = paths.iterator(); it.hasNext();) {
+            String path = it.next();
+            if (path.equals("commits"))
+            	break;
+            it.remove();
+        }
 
         String token = req.getParameter("token");
 


src/main/java/com/dabsquared/gitlabjenkins/GitLabWebHook.java
```diff
@@ -108,6 +108,14 @@ public void run() {
         while (restOfPathParts.hasNext()) {
             paths.add(restOfPathParts.next());
         }
+        
+        // remove everything till we found 'commits'
+        for (Iterator<String> it = paths.iterator(); it.hasNext();) {
+            String path = it.next();
+            if (path.equals("commits"))
```
xhaggi: after a second look I've missed that the path could also hold "build" and "builds". I'll create a new PR with a better approach tomorrow.

djb61:
I did a similar but not quite the same fix for myself locally to work around this issue. I didn't raise a pull request however as I didn't think it was enough to just mimic the previous behaviour.

Now that Gitlab is providing a reference to the source branch should we not also validate that, so rather than just finding a build with the correct commit hash find a build which both has the correct commit and was built from the specified source branch?

xhaggi:
@djb61 for me it's enough to fix the issue so that we can use it with the latest version of GitLab. Everything else is more work and should be done by someone who knows the code base a little bit better than me :smile:

PR #83, opened 2015-05-24
Gitlab action type env var

sullb:
Adding new environment variable to indicate if trigger was due to PUSH or MERGE.

This will allow for the use of conditional build step to change the behaviour of the build.

If this is possible another way, please let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ These include:
 * gitlabSourceRepoURL
 * gitlabSourceRepoName
 * gitlabBranch (This is optional and can be used in shell scripts for the branch being built by the push request)
-* gitlabActionType (This can be either PUSH or MERGE)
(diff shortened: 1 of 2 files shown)


bassrock:
@sullb Does this require a change to the setup process? If so please add it to the realm and add tot he pull request.

sullb:
@bassrock setup should be the same, just an extra jenkins variable (${gitlabActionType}) is available for other plugins to use, I have added the extra parameter under Parameterized builds in README.md.

If you mean something else could you explain further please.

bassrock:
Ok, and is the parameter required to add to the job or is it optional to add it?

sullb:
@bassrock good point, I have clarified that it is optional in README.md

PR #100, opened 2015-08-20
add support for workflow projects

sudssm:
The goal of this branch is to support GitLab Push Triggers in the Workflow project type.

The current state of this branch works for Workflow builds!

collaborating with @omehegan and @lukeknep


sudssm:
@bassrock requesting comment

PR #105, opened 2015-09-02
JENKINS-28271 - Gitlab Plugin does not build Merge Request when first raised

jr4:
prevent NPE that was preventing the build from proceeding
tested that the build proceeds fine after that


skyline-gleb:
@bassrock +1

WonderCsabo:
I can also verify this fix works well.

kleijnweb:
Running into this sloppyness with older version of GitLab as well. 

LGTM +1.

WonderCsabo:
We are using an old version (GitLab 6.9) unfortunately. BTW, is this change working with the latest GitLab versions as well?

kleijnweb:
It should. Looks compatible with examples from latest GitLab docs.

Issue #120, opened 2015-10-16
Checkbox acceptMergeRequestOnSuccess set to "checked" always


trunov-ms:
When I open Jenkins Job Configure page, acceptMergeRequestOnSuccess checkbox set to "checked" state always even if there was unchecked earlier and option <b>acceptMergeRequestOnSuccess</b> in file Jobname/config.xml set to <b>false</b> too.

gumallett:
+1

jsyrjala:
I made a related issue to jenkins jira: https://issues.jenkins-ci.org/browse/JENKINS-31122. @kasper-f

nerzhul:
The XML was okay, but everytime the plugin merge requests, it's anoying

PR #123, opened 2015-10-25
Fix for issue #120

tomelfring:
There is a getter needed so you can update variables.


omehegan:
@thommy101 I think I prefer the diff in https://github.com/jenkinsci/gitlab-plugin/pull/124 because it also changes the default for this setting to 'false,' which I think is important. If you agree, let's close this PR and I'll merge that one. Thanks!

tomelfring:
@omehegan I'll agree with that, setting it to false is indeed a very good idea.

PR #129, opened 2015-11-06
Assorted minor fixes for merge requests

TomAndrews:
-  Make ci-skip for merge requests respect the configuration
-  Use new API call to get merge requests
-  Logging fix

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/GitLabWebHook.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/GitLabWebHook.java
@@ -417,8 +417,8 @@ public class GitLabWebHook implements UnprotectedRootAction {
 									+ mr.getSourceBranch() + "\n target: "
 									+ mr.getTargetBranch() + "\n state: "
 									+ mr.getState() + "\n assign: "
-									+ mr.getAssignee() + "\n author: "
-									+ mr.getAuthor() + "\n id: "
+									+ mr.getAssignee().getName() + "\n author: "
+									+ mr.getAuthor().getName() + "\n id: "
 									+ mr.getId() + "\n iid: "
                                     + mr.getIid() + "\n last commit: "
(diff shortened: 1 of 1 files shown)


omehegan:
@TomAndrews thanks! Can you:
1. Give some more detail about the new API call you're using? I just want to make sure we don't break backwards compatibility when we do new releases of the plugin.
2. Resolve git conflicts (maybe just rebase).

TomAndrews:
Thanks for taking a look.  I have rebased and fixed the conflict.

The API change is in commit 8eedf3dfd66f2f78b4f93f235fb95d1acbf461d6.

Before the code was using

```
String tailUrl = GitlabProject.URL + "/" + projectId + GitlabMergeRequest.URL + "?state=opened&per_page=100";
List<GitlabMergeRequest> mergeRequests = api.instance().retrieve().getAll(tailUrl, GitlabMergeRequest[].class);
```

I am proposing that we switch to the `getOpenMergeRequests` call defined [here](https://github.com/timols/java-gitlab-api/blob/b8fc0e6b08d5ca6cc1f85fd185e1b0606a599db8/src/main/java/org/gitlab/api/GitlabAPI.java#L628).

The URL being fetched is almost identical:

```
String tailUrl = GitlabProject.URL + "/" + projectId + GitlabMergeRequest.URL + "?state=opened&per_page=100";
```

vs 

```
String tailUrl = GitlabProject.URL + "/" + project.getId() + GitlabMergeRequest.URL + "?state=opened";
```

The API call just misses out the parameter `per_page=100`.  This was added in fc493523b092f1cbbd86efb7388c8ed0aae2a416 along with the note to switch to the getOpenMergeRequests call once API v 1.1.7 was released. I guess it shouldn't cause any problems but I'm no expert!

omehegan:
Thanks!

PR #141, opened 2015-11-22
Refactoring data objects to own package

theobisproject:
- Moved data objects like Branch to own data package
- Using [lombok](https://projectlombok.org/) to generate getter and setter

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/GitLabMergeRequest.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/GitLabMergeRequest.java
@@ -1,15 +1,14 @@
 package com.dabsquared.gitlabjenkins;
 
-import java.io.IOException;
-import java.util.Date;
-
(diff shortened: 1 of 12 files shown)


omehegan:
@markus-mnm would you mind taking a look at this when you have a chance? It's a big change and well above my expertise to assess whether we should take it on board :)

markus-mnm:
I'll check it out this weekend.

markus-mnm:
I believe moving those pojos out is a good idea.
On the other hand I am against using lombok, especially since it requires an IDE plugin.
I would then rather not use getter/setters at all.

EDIT: removed down-vote since that doesn’t seem to do in github what I thought.

theobisproject:
So should i rebase it against master and throw away the lombok commit?

markus-mnm:
yes, please

omehegan:
@theobisproject apologies for the long silence. Is this good to merge now, minus the lombok stuff?

theobisproject:
Yes, lombok stuff is not contained

PR #154, opened 2015-12-10
8.1 is now a supported version

dosire:


Conan-Kudo:
@dosire Does this not work with GitLab 8.2.x too?

omehegan:
@Conan-Kudo I have not heard any reports about GitLab 8.2, if you can verify that I'd be happy to know the result!

dosire:
@Conan-Kudo should work in 8.2 too, no changes in the commit status api as far as I know

Issue #160, opened 2015-12-17
Merge request does not trigger build because it has been merged already


AndreasBoehme:
1. create a Merge request for branch "feature_1" to "release_1" - Jenkins merges the branch. OK.
2. create a Merge request for branch "feature_1" to "master" - the gitlab plugin silently denies that one because: "_Last commit in Merge Request has already been built in build #1234_"

coder-hugo:
Close this as #163 is merged back

PR #163, opened 2015-12-17
Bug #160: allow merge requests to different branches from one commit.

AndreasBoehme:
Fixes my issue #160.
Take a look at GitLabWebHook.java, starting at line 483 (the rest is just formatting).

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/GitLabWebHook.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/GitLabWebHook.java
@@ -1,6 +1,7 @@
 package com.dabsquared.gitlabjenkins;
 
 import com.dabsquared.gitlabjenkins.GitLabMergeRequest.LastCommit;
+
 import hudson.Extension;
 import hudson.model.*;
 import hudson.plugins.git.Branch;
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -7,7 +7,7 @@
     </parent>
 
   <artifactId>gitlab-plugin</artifactId>
-  <version>1.1.29-SNAPSHOT</version>
+  <version>1.1.29.2-SNAPSHOT</version>
```
omehegan: No need for this change. When we do a release, the plugin version will increment automatically. So please revert this :)

omehegan:
@AndreasBoehme looks like a good change, thanks! As much as I want to clean up the formatting of this plugin, I would like to do it in one big merge rather than in small steps. That way we can minimize noise in diffs. I'm planning to work on this after I get back from vacation. Would it be very painful for you to resubmit this PR as _just_ the functional change without reformatting? If so, I'll take it as-is. Thanks!

AndreasBoehme:
I reverted the previous changes and applied only the essential ones. (it is a push -force - hope that's ok)

omehegan:
@AndreasBoehme thanks, this looks good! I will plan to do a release after I get back from vacation in mid-January.

In terms of the formatting changes, I would like to establish a standard style for the project and then use a tool such as checkstyle to enforce it. I just need to do a little research into how best to accomplish this. Maybe I can get that done before the next release.

ddwolf:
what if I just want to rebuild this commit it while the source commitid and target branch not change?
e.g. after I changed by jenkins job configuration

omehegan:
@ddwolf you would have to trigger the Jenkins build manually. Could be more or less challenging depending on the project type (relatively easy with Pipeline Multibranch, other project types would need a parameter defined to let you specify the repo branch for manual builds).

PR #165, opened 2015-12-22
Merge requests should also set gitlabBranch

razzeee:
Not sure I understand why 'gitlabBranch' is there, when I could also use one of the other parameters fine in Jenkins.
So we should either deprecate it or also set it on merge requests. Like this.

(this is untested)


omehegan:
@Razzeee I don't understand the purpose of this change. It's adding a parameter named gitlabBranch whose value is identical to gitlabSourceBranch?

razzeee:
Well push requests have this. So we should align this to them. Or remove it from both. 
The parameter is also documented here https://github.com/jenkinsci/gitlab-plugin/blob/master/README.md#parameterized-builds

coder-hugo:
Will be fixed by #211 which replaces all currently available build parameters by build variables (and adds also some missing ones).

Issue #164, opened 2015-12-21
Not reporting build status 8.1+


razzeee:
Hey there,
this is probably not a bug, but a user problem / bad documentation.

Since we upgraded to 8.2.3 I can't get gitlab to speak to jenkins. And the readme is not that helpful, do I need to enable Gitlab CI? This features looks completely different in gitlab 8+
CI Web Hooks are probably not used?

So I figured to disable it and use the normal Web Hooks, as it were before gitlab 8, but the build status doesn't update either way.
Also pushes seem to just crash the build due to git problems. Weird thing is, merge requests initial push does build fine.

Gitlab 8.2.3
Jenkins 1.642
All jenkins plugins are up to date

razzeee:
On the pushes crashing, it seems to be due to 'Merge before build'
As ${gitlabTargetBranch} seems to be equal to my ${gitlabSourceBranch}
Seems weird, as this only happens on pushes. Might be a gitlab issue :/

razzeee:
That code seems weird.
https://github.com/jenkinsci/gitlab-plugin/blob/258f65b8900fa7a9d73ee202101a3d58da6856d9/src/main/java/com/dabsquared/gitlabjenkins/GitLabPushTrigger.java#L237
Shouldn't that get the correct values, as the merge request does?

PR #166, opened 2015-12-22
Push requests should also build against the target branch

razzeee:
Before this jenkins will try to merge our source branch with a branch of the same name on our upstream git server. Which mostly means it will fail due to that branch not existing. We should use the branch name the gitlab push request send us.
Also see: #164

(untested so far, but the jenkins build triggered by this PR might help)


razzeee:
I just realized, that the gitlab hook isn't exposing the needed data. So we can't fix building push requests with merging the upstream.
Sorry, for the noise.

PR #169, opened 2016-01-02
fix null pointer access in buildOpenMergeRequests

runsisi:
assignee is not a mandatory field for a merge request, we need to test it before accessing its members

Signed-off-by: runsisi runsisi@hust.edu.cn


WonderCsabo:
I can verify this PR is needed and works. However i think author is mandatory, only assigne is not.

coder-hugo:
Was the same as #195

PR #188, opened 2016-01-30
[WIP] Reordered attributes and methods

theobisproject:
I've done some reordering to correspond the java standards. While doing it I had the feeling all attributes in GitLabPushRequest except sourceProject should be extracted to an extra object but I couldn't come up with a good name for it.
Could someone tell me for what the object_kind in GitLabMergeRequest is needed?


coder-hugo:
Thanks for your work, but I'll do a bigger code cleanup which will contain this changes in a similar way.

PR #192, opened 2016-02-12
Add publishers to update the commit/merge request status in GitLab

### LINKED ISSUE #162: Build status for first push to new branch does not get reported to GitLab
If you create a new branch and push some changes, a build is triggered. However, no commit status is reported to GitLab. On subsequent commits/pushes to that branch the status is reported properly.

coder-hugo:
This PR moves the configuration (and execution) of the GitLab commit status updates to a publisher ("Publish build status to GitLab commit (GitLab 8.1+ required)"). With this change its possible to trigger builds manually without adding any special parameter and the status of the corresponding commit in GitLab gets updated. Additionally there is a XStream converter that migrates the configuration of old jobs to use this publisher automatically depending on their GitLabPushTrigger's configuration.
There is just one thing missing of the previous functionality: The commit status "pending" will not be sent to GitLab. The reason for this is that I haven't found any extension point that throws an event or something similar if a build was triggered but not yet started.
Fixes #162


coder-hugo:
Close this as there is no possibility to change the target branch. New PR #229

PR #197, opened 2016-02-19
Reorganize README to be more clear about config for different versions of GitLab

omehegan:


omorillo:
Solution to this issue is underway.

PR #207, opened 2016-02-23
Cleanup GitLabWebHook

coder-hugo:
This PR cleans the code of the GitLabWebHook by splitting it into multiple classes.
The PR is based on #198.

Issue #90, opened 2015-06-30
Build on push for existing merge request is not working


jlove4m:
Hi!

We're having an issue with this plugin. What we're seeing is that when someone does a push for an open merge request a build is not triggered. Regular merge requests are, however, being built. If we enable build on pushes they also work - we're just having an issue with pushes on open merge requests.

Here's what the System log looks like in jenkins:

Jun 30, 2015 2:37:51 PM INFO com.dabsquared.gitlabjenkins.GitLabWebHook getDynamic
WebHook called with url: /commits/ccd968415cae32c93a4136b24235082f8de44028/status.json
Jun 30, 2015 2:45:58 PM INFO com.dabsquared.gitlabjenkins.GitLabWebHook getDynamic
WebHook called with url: /build
Jun 30, 2015 2:45:58 PM WARNING org.eclipse.jetty.util.log.JavaUtilLog warn
Error while serving https://ci2.thorleyindustries.com/project/AMS-TestJenkinsIntegration/build
java.lang.reflect.InvocationTargetException
	at sun.reflect.GeneratedMethodAccessor284.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.kohsuke.stapler.Function$InstanceFunction.invoke(Function.java:298)
	at org.kohsuke.stapler.Function.bindAndInvoke(Function.java:161)
	at org.kohsuke.stapler.MetaClass$13.dispatch(MetaClass.java:409)
	at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:746)
	at org.kohsuke.stapler.Stapler.invoke(Stapler.java:876)
	at org.kohsuke.stapler.MetaClass$13.dispatch(MetaClass.java:411)
	at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:746)
	at org.kohsuke.stapler.Stapler.invoke(Stapler.java:876)
	at org.kohsuke.stapler.Stapler.invoke(Stapler.java:649)
	at org.kohsuke.stapler.Stapler.service(Stapler.java:238)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:848)
	at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:686)
[...]

PR #208, opened 2016-02-24
Allow build to be triggered when pushing to an Open MR

daddykotex:
The function generateMergeRequestBuild return when it should not be if the state
equals to "update". I removed the check to allow the build to be triggered.

I also used StringUtils.isEmpty to perform a check in GitLabWebHook to avoid a NPE.


jtheoof:
Looks good, seems to be fixing #90 Works fine with our dev environment.

coder-hugo:
Open MRs will be triggered already by pushing to the corresponding branch if the option "Rebuild open Merge Requests" is set at least to "On push to source branch" (Tested with GitLab 7.14.3 and GitLab 8.4.3 using the code of PR #207 which should mainly contain code cleanups and no bug fixes)
If we use the update Merge Request Event to trigger builds for a open MR (which IMHO is the cleaner way as we don't need to figure out the MR for an incoming Push Event) we should remove the code that triggers a new MR build in case of a Push Event.

daddykotex:
Looks like @coder-hugo did all the work in #207 , thank you for that!

PR #171, opened 2016-01-02
fix error "failed to communicate with gitlab server ..." when try to rebuild open MRs

runsisi:
as stated in [comment](https://github.com/jenkinsci/gitlab-plugin/issues/144#issuecomment-167432522) of #144,  when we try to rebuild opened MRs, we should try to get the source branch by the id of source project, or else we will meet this:

```
failed to communicate with gitlab server to determine is this is an update for a merge request: http://hust/api/v3/projects/4/repository/branches/mr4?private_token=UmK47w8rxu7PfA3Fuu6-
```

Signed-off-by: runsisi runsisi@hust.edu.cn

PR #215, opened 2016-03-05
Fix error "failed to communicate with gitlab server ..." when try to rebuild open MRs

WonderCsabo:
Same as #171, only the formatting changes were removed from the commit.


WonderCsabo:
@coder-hugo @omehegan can you check this PR? This is needed for newer Gitlab versions, at least from 8.3.

PR #173, opened 2016-01-05
Configure Gitlab URL at job level

WonderCsabo:
Implements https://github.com/jenkinsci/gitlab-plugin/issues/121.

This is my first contribution to a Jenkins plugin, so please review carefully. Thanks!

Issue #121, opened 2015-10-16
Configure GitLab host url and API token at job level


WonderCsabo:
Currently the GitLab host url and API token are configured globally. This means one Jenkins instance can only communicate to one, specific GitLab instance. It would be nice if we could override this setting at job level.

mattiasb:
I'd propose to store this with the Credentials plugin if possible. So you could just have a drop down menu to choose what Gitlab instance to use.

WonderCsabo:
Good idea. However i would be really glad for any solution which would move this setting to job level. :smile:

omehegan:
https://github.com/jenkinsci/gitlab-plugin/pull/198 will resolve this I think.

PR #230, opened 2016-03-10
Add functionality to configure multiple gitlab connections

coder-hugo:
Implement #121 and makes PR #173 obsolete.
This PR adds the functionality to configure multiple gitlab instances in the global settings. The job configuration gets an additional property to configure the gitlab connection for this job.
The PR is based on #229 as it needs some of the refactorings done in this PR.


WonderCsabo:
@coder-hugo we just tried this PR, but we observe several problems :cry: 
- pushing a new commit on an existing MR does not trigger a build
- the build status cannot be seen in Gitlab at all
- config will have unique name validations on already existing Gitlab connection names
- the ignoreHttpsCertError setting seem to be forgotten

coder-hugo:
@WonderCsabo 
- pushing a new commit on an existing MR does not trigger a build => **works for me** (GitLab 7.14 and 8.5)
- the build status cannot be seen in Gitlab at all => I see the build status in GitLab 8.5, for GitLab 7.14 there seems to be something broken but not by this change. I get always the pending status also without this PR. The Jenkins comments to the MR are available in both GitLab 7.14 and 8.5.
- config will have unique name validations on already existing Gitlab connection names => **fixed**
- the ignoreHttpsCertError setting seem to be forgotten => **fixed** (it wasn't forgotten just not displayed due to an error in the config.jelly)

WonderCsabo:
Thanks! We updated the plugin from the previous version. Did you try that as well?

coder-hugo:
@WonderCsabo: yes I configured the gitlab-plugin and a job using the previous version and updated the plugin afterwards to this branch. After the update everything works as expected.

WonderCsabo:
I see. I do not know what went wrong in our Jenkins, then. :( Is there a way to debug a normal Jenkins instance?

coder-hugo:
@WonderCsabo: If you really want to debug a "normal" Jenkins instance you can start it by passing the Java remote debugging arguments to the JRE using an environment variable JAVA_OPTS or something similar. If the Jenkins JVM runs in debug mode you can connect to it using the remote debugging of your favorite Java IDE. But I wouldn't suggest you to do this on a productive Jenkins which is used by many others cause the debugging mode slows the JVM down.
I think you should first try to add a logger for "com.dabsquared.gitlabjenkins" using at least the log level FINE to get more information what went wrong.

WonderCsabo:
@coder-hugo thanks! Actually updating this plugin to the new version kills all the MR jobs, so it is not a problem if Jenkins is slow. :) I will try the logging first, though.

PR #231, opened 2016-03-10
Cleanup GitLabWebHook

coder-hugo:
This PR cleans the code of the GitLabWebHook by splitting it into multiple classes.
The PR is based on #230


markus-mnm:
I had a look and like what I saw so far. I will try to do some testing tomorrow.  Caveat: I have not been the master of my schedule for some time and I would not wait for my OK on this.

Regarding the test JSONS, I had a similar approach that I used locally. This PR's has more complete data, but I also used jsons from different versions of Gitlab to test some of the peculiarities of different versions. 

Another thing I noticed so far is that some of the debug and info logging is gone. E.g. the request payload and such used to be useful to find problems with different versions.  I still have to run this PR's  code, so this might be a useless observation.

PR #244, opened 2016-03-22
Add setters for all parameters of the Gitlab system config to the trigger descriptor

emmanuel099:
This makes it possible to change the Gitlab system config (host url, api token and ignore ssl errors) via a Groovy-script, e.g.:

init.groovy:

```
import jenkins.model.*

def jenkins = Jenkins.getInstance()
...

def gitlab = jenkins.getDescriptor("com.dabsquared.gitlabjenkins.GitLabPushTrigger")
gitlab.setGitlabHostUrl("https://url-to-gitlab")
gitlab.setGitlabApiToken("some-gitlab-token")
gitlab.setIgnoreCertificateErrors(true)
gitlab.save()
```

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/GitLabPushTrigger.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/GitLabPushTrigger.java
@@ -938,12 +938,24 @@ public class GitLabPushTrigger extends Trigger<Job<?, ?>> {
             return gitlabApiToken;
         }
 
+        public void setGitlabApiToken(String gitlabApiToken) {
+            this.gitlabApiToken = gitlabApiToken;
+        }
+
         public String getGitlabHostUrl() {
             return gitlabHostUrl;
         }
 
+        public void setGitlabHostUrl(String gitlabHostUrl) {
+            this.gitlabHostUrl = gitlabHostUrl;
+        }
+
         public boolean getIgnoreCertificateErrors() {
         	return ignoreCertificateErrors;
+        }
+
+        public void setIgnoreCertificateErrors(boolean ignoreCertificateErrors) {
+            this.ignoreCertificateErrors = ignoreCertificateErrors;
         }
 
         public static DescriptorImpl get() {


coder-hugo:
Changing these settings within a script makes no sense for me. So one job could break several other jobs by changing these settings.
I assume that #121 is a better solution for what you are looking for.

emmanuel099:
> I assume that #121 is a better solution for what you are looking for.

No, 121 is about overriding the systems settings in each job.

This pull request is about changing the systems settings in Jenkins init scripts [1], not in job scripts.

[1] https://wiki.jenkins-ci.org/display/JENKINS/Post-initialization+script

emmanuel099:
I use this init script (short version) in a Jenkins docker container:

```
import jenkins.model.*

def jenkins = Jenkins.getInstance()
jenkins.setNumExecutors(8)
jenkins.setScmCheckoutRetryCount(3)
jenkins.save()

def locationConfiguration = JenkinsLocationConfiguration.get()
locationConfiguration.setUrl("https://jenkins:8080/")
locationConfiguration.save()

def mailer = jenkins.getDescriptor("hudson.tasks.Mailer")
mailer.setSmtpAuth("user", "password")
mailer.setSmtpHost("host")
mailer.setUseSsl(true)
mailer.save()

def gitlab = jenkins.getDescriptor("com.dabsquared.gitlabjenkins.GitLabPushTrigger")
gitlab.setGitlabHostUrl("https://gitlab")
gitlab.setGitlabApiToken("asdf")
gitlab.setIgnoreCertificateErrors(false)
gitlab.save()
```

The Gitlab API token should later be injected by a environment variable, so that I can simply change the API token via "docker run --env". The Jenkins container is used by different people, so I can't set a static API token in the XML config file.

coder-hugo:
This will not work starting with version 1.2 (planned to be released in the next few days).
This will add the GitLabConnectionConfig which stores all required GitLab connections. With this you could use the addConnection method in your script for configuring GitLab.

emmanuel099:
> This will add the GitLabConnectionConfig which stores all required GitLab connections.

Ok missed that somehow (not merged yet?), but sounds really good :)
Thanks!

PR #245, opened 2016-03-23
feat: GitLab OAuth2 provider as Jenkins SecurityRealm

bufferoverflow:
Signed-off-by: Roger Meier r.meier@siemens.com
Reviewed-by: Fabio Huser fabio.huser@siemens.com
Tested-by: Rudolf Honegger rudolf.honegger@siemens.com

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/gitlab/GitlabSecurityRealm.java
+++ b/src/main/java/org/jenkinsci/plugins/gitlab/GitlabSecurityRealm.java
@@ -1,0 +1,223 @@
+// Copyright (C) 2016, Siemens AG
+// SPDX-License-Identifier:	GPL-2
+
+package org.jenkinsci.plugins.gitlab;
+
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.UnsupportedEncodingException;
+import java.net.URL;
+import java.net.URLConnection;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import org.acegisecurity.Authentication;
+import org.acegisecurity.AuthenticationException;
+import org.acegisecurity.AuthenticationManager;
+import org.acegisecurity.BadCredentialsException;
+import org.acegisecurity.GrantedAuthority;
+import org.acegisecurity.context.SecurityContextHolder;
+import org.acegisecurity.providers.UsernamePasswordAuthenticationToken;
+import org.acegisecurity.providers.anonymous.AnonymousAuthenticationToken;
+import org.gitlab.api.GitlabAPI;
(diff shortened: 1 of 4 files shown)


omorillo:
Hello @bufferoverflow, Thank you for your pull request. I will review the code and test the feature and if everything works fine I will merge it.

omehegan:
@bufferoverflow can you explain what this feature enables? When you refer to the Jenkins SecurityRealm, it makes me think this allows you to authorize access to Jenkins based on accounts in GitLab. If that's correct, I'm not sure it makes sense to include this in our plugin, which is currently just a build trigger interface. 

Is there any reason why the generic Jenkins oauth plugin can't be used to authorize against a GitLab system?

I apologize if I've just misunderstood the intent of this change.

omorillo:
No, I think you're right @omehegan. It seems to be rather some Jenkins core functionality. Me too was wondering about this sort of feature for our gitlab-plugin and wanted to look what is the functionality about to check whether I am wrong with my understanding. But I hope @bufferoverflow can bring some clarification to this question.

bufferoverflow:
Yes, this extension allows Authorization of Jenkins users via GitLab. 

OAuth is not that generic as we would like to have it, each OAuth service provider has little different semantics, there are a lot of different OAuth plugins around e.g. Google, GitHub.

We decided to create a pull request here, because the plugin name is generic **gitlab-plugin** and we prefer to have one plugin with GitLab functionality instead of adding an additional plugin. I have the feeling there might be other synergies in the future.

omorillo:
@bufferoverflow The name of the plugin is pretty generic, this for sure. Still this feature deviates completely from the idea of the gitlab-plugin which is to trigger jobs on push to a certain branch or on submission of a merge request. I don't think this feature fits to this plugin.

omorillo:
Do you see it similar, @omehegan?

bufferoverflow:
I would appreciate I you could give it a try, it does not influence any functionality of your plugin. It's just an addon and I believe it is a worth to have addon for GitLab users that are searching for GitLab plugins.

omehegan:
Let me take a look and think about it, I was involved with some work on the Jenkins OAuth plugins a year or two ago. At a minimum I think people would not expect to find this auth functionality inside the GitLab plugin, so if we did adopt it we would need to change the way we present it.

bufferoverflow:
And of course, we are using your plugin already.

omehegan:
Upon discussion amongst ourselves (the maintainer team) and with core Jenkins developers, we agree that the best practice would be for this feature to be implemented as its own plugin. This pattern is used in various similar cases, such as the separate GitHub and GitHub auth plugins. As such, while we really appreciate that you submitted this feature, we think this is not the best place for it. If you need help getting this refactored to be its own plugin, let us know and we can point you to some resources for that. Thank you!

Issue #250, opened 2016-03-29
Intermittent failures to trigger a Jenkins build when an MR first opened


cotman:
Hi,

I've noticed comments in other issues, elsewhere, but they seem to suggest they've gone away?

I'm running:

Jenkins: 1.654
Gitlab Plugin: 1.1.32
Gitlab: 8.5.4

For the most part, everything works brilliantly.  However...occasionally when opening an MR, a build isn't triggered and I see the following in the `jenkins.log`:

```
Mar 29, 2016 2:56:15 PM com.dabsquared.gitlabjenkins.GitLabWebHook getDynamic
INFO: WebHook called with url: 
Mar 29, 2016 2:56:15 PM org.eclipse.jetty.util.log.JavaUtilLog warn
WARNING: Error while serving https://-removed-url-/project/-removed-project-name
java.lang.reflect.InvocationTargetException
at sun.reflect.GeneratedMethodAccessor1121.invoke(Unknown Source)
at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
at java.lang.reflect.Method.invoke(Method.java:606)
at org.kohsuke.stapler.Function$InstanceFunction.invoke(Function.java:324)
at org.kohsuke.stapler.Function.bindAndInvoke(Function.java:167)
at org.kohsuke.stapler.MetaClass$11.dispatch(MetaClass.java:378)
at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:746)
at org.kohsuke.stapler.Stapler.invoke(Stapler.java:876)
at org.kohsuke.stapler.MetaClass$11.dispatch(MetaClass.java:380)
at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:746)
at org.kohsuke.stapler.Stapler.invoke(Stapler.java:876)
at org.kohsuke.stapler.Stapler.invoke(Stapler.java:649)
at org.kohsuke.stapler.Stapler.service(Stapler.java:238)
at javax.servlet.http.HttpServlet.service(HttpServlet.java:848)
at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:686)
at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1494)
at hudson.util.PluginServletFilter$1.doFilter(PluginServletFilter.java:134)
[...]

Issue #249, opened 2016-03-29
Two or more webhooks produce 'retried builds' in GitLab


vills:
_Not sure, but it can be a duplicate of #200_

If we use more than one webhook in repository, than builds are recorded in GitLab as 'retried', that isn't true actually. Not sure, but it can be because of field 'Name' (Jenkins) reported by plugin to GitLab instance.

It would be nice to set-up 'Name' field on per job basis in jenkins, that is reported to GitLabCI.

Gitlab version: 8.6. Gitlab 8.1 features enabled in jenkins' job.

Thanks.

coder-hugo:
IIRC it's enough if the description of the [commit status](http://doc.gitlab.com/ce/api/commits.html#post-the-build-status-to-a-commit) is different to have a separate build instead of a retried one in the GitLab build list.
So I think we should add the possibility to send a also configurable description to GitLab while updating the status of a commit.

vills:
@coder-hugo +1

Anyway, it would be nice to see both 'name' and 'description' as a optional configurable options in job. Cause 'name' add ability to determine job exactly from gitlab's build list.

coder-hugo:
Duplicate of #362

Issue #239, opened 2016-03-18
Improve handling of Merge Requests


pbendersky:
Based on a question I asked on GitLab (see [here](https://gitlab.com/gitlab-org/gitlab-ce/issues/14181#note_4212485)), it turns out we can fetch the contents of a Merge Request from the `origin` repository. As I see it, this has several advantages:
1. You don't need to setup your forked repositories to use CI, nor set the deploy key to them.
2. We could configure GitLab projects with a single repository (and maybe with no parameters!).
3. As we'll have a single repository, the "Build" button in Jenkins will remain functional.

I've only reviewed the source code quickly, so I don't have a deep understanding of it (yet). Does anyone see a drawback of this? Is this an improvement that would be welcome? (I understand this might break for older versions of GitLab).

omehegan:
Sounds pretty cool... we should look into it further. We especially need to understand what versions of GitLab support this.

pbendersky:
Is there a minimum compatibility requirement? Or we can break the plugin for older versions of GitLab?

omehegan:
GitLab changed a lot between version 7.14 and 8.1. As a result, the plugin has a "Use GitLab 8.1 features" check box in the config. If this feature is only available in 8.1+, we could enable it along with the other new features based on that check box status. IMO we need to support at least 7.14, it seems many people are still using that, including me :P

pbendersky:
Here's the full documentation of these refs: http://doc.gitlab.com/ce/workflow/merge_requests.html

This was added in [8.0.0](https://gitlab.com/gitlab-org/gitlab-ce/blob/c790107854658a37d97b9a1c9066338eb12b06bf/CHANGELOG#L662), so it's safe to use it with the 8.1+ flag.

omehegan:
@pbendersky you figured out a way to do this with just different Git configuration in Jenkins, right? No code changes needed? If this is ready to close, go ahead and do that. Thanks!

pbendersky:
Yes. I still think there should be a better approach (with a single repo configuration in Jenkins). But I couldn't made it work this time. I'll try sometime else.

Issue #238, opened 2016-03-18
New branch push causes error in jenkins


rhduwe:
Pushes in existing branches trigger builds normally. When I push a new branch, build is not triggered and in logs there is a warning.

UPD: Problem seems to have gone away after updating gitlab to 8.5.8

Jenkins versions I've tried: 1.653, 1.641
Gitlab versions:  8.3.4, 8.5.7

Jenkins configuration - completely empty with git and gitlab plugins installed. 
Project - configured git repository (gitlab ssh), enabled 'Build when a change is pushed to GitLab', added simple echo shell script as a build step.
Gitlab - created project (without .gitlab-ci.yml), added webhook to http://jenkins:8080/project/testproject, created a user for jenkins, added public key

`Error while serving http://jenkins:8080/project/testproject
java.lang.reflect.InvocationTargetException
    at sun.reflect.GeneratedMethodAccessor387.invoke(Unknown Source)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:606)
    at org.kohsuke.stapler.Function$InstanceFunction.invoke(Function.java:324)
    at org.kohsuke.stapler.Function.bindAndInvoke(Function.java:167)
    at org.kohsuke.stapler.MetaClass$11.dispatch(MetaClass.java:378)
    at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:746)
    at org.kohsuke.stapler.Stapler.invoke(Stapler.java:876)
    at org.kohsuke.stapler.MetaClass$11.dispatch(MetaClass.java:380)
    at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:746)
    at org.kohsuke.stapler.Stapler.invoke(Stapler.java:876)
    at org.kohsuke.stapler.Stapler.invoke(Stapler.java:649)
    at org.kohsuke.stapler.Stapler.service(Stapler.java:238)
    at javax.servlet.http.HttpServlet.service(HttpServlet.java:848)
    at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:686)
    at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1494)
    at hudson.util.PluginServletFilter$1.doFilter(PluginServletFilter.java:134)
[...]

PR #252, opened 2016-03-30
Plugin modes for better MR handling.

pbendersky:
This PR enhances gitlab-plugin by adding two operation modes: Modern and Legacy.
**Legacy** is what we currently have, that works with older versions of GitLab.

**Modern** works with GitLab 8.1 or better, and provides an easier configuration of projects in Jenkins (a single Git repository is enough) and much better handling of Merge Requests (MR are fetched from `origin` and matched by SHA1 instead of comparing branch names).

This PR fixes the two issues I reported: #239 and #155.
Also, a quick review of the current issues indicates that it should fix (assuming they can use the Modern mode):
- #250, #238 and #210 (seem related to MR handling)
- #249, #194 (since builds are matched by SHA1, duplicates are prevented easily)
- #232 (since it won't use the title to construct the git URL)
- #136, #98 

This has been implemented with a strategy pattern. The points where the strategy is used are ad-hoc based on where the Legacy and Modern strategies differ (a few but important points).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,22 @@
 # Table of Contents
 - [Introduction](#introduction)
 - [User support](#user-support)
-- [Supported GitLab versions](#supported-gitLab-versions)
+- [Plugin Modes and Supported GitLab versions](#plugin-modes-and-supported=gitlab-versions)
 - [Supported GitLabCI Functions](#supported-gitlabci-functions)
 - [Configuring access to GitLab](#configuring-access-to-gitlab)
+- [Modern Mode](#modern-mode)
     - [Jenkins Job Configuration](#jenkins-job-configuration)
-    - [Gitlab Configuration (7.14.x)](#gitlab-configuration-714x)
     - [Gitlab Configuration (>= 8.1)](#gitlab-configuration--81)
+    - [Manual Trigger](#manual-trigger)
+- [Legacy Mode](#legacy-mode)
+    - [Jenkins Job Configuration](#jenkins-job-configuration-1)
+    - [Gitlab Configuration (7.14.x)](#gitlab-configuration-714x)
+    - [Gitlab Configuration (>= 8.1)](#gitlab-configuration--81-1)
     - [Forked repositories](#forked-repositories)
-- [Branch filtering](#branch-filtering)
-- [Build Tags](#build-tags)
-- [Parameterized builds](#parameterized-builds)
+- [Branch filtering (Modern & Legacy)](#branch-filtering-modern--legacy)
+- [Build Tags (Modern & Legacy)](#build-tags-modern--legacy)
 - [Help Needed](#help-needed)
+- [Contributing to the Plugin](#contributing-to-the-plugin)
 - [Quick test environment setup using Docker](#quick-test-environment-setup-using-docker)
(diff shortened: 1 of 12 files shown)


cotman:
This looks massively interesting to us, as we're impacted by one of the issues you refer to.  A couple of questions, if that's OK (and this is the appropriate place for this):
- Will this still play well with jobs that take a branch as a _parameter_ now?  As an example, we currently have single jobs that can be run manually by developers by launching the job in Jenkins and setting the `gitlabSourceBranch` parameter to the branch they want to test _and_ that are also triggered via Gitlab webhooks automagically using the current Gitlab Plugin (which sets `gitlabSourceBranch`).  If I'm reading this correctly then a default, empty, branch parameter (actual name of the parameter now wouldn't matter) would work for both cases using Modern mode i.e. developer populates the branch parameter for a manual build (unpopulated by default - the job is configured so that the Git plugin sets "Branches to build" based on it), webhook triggered builds have the branch deduced by the plugin, and work on the correct branch as a result?
- Not totally related, but as the README.md has been updated - is it worth clarifying, a bit more, what "Use GitLab CI features" means with regard to how the Gitlab Plugin works?  It might be just be me, but I've always been a little confused as to everything it offers (we always have it enabled).  It appears under the Legacy mode with the updated README.md, so is no longer relevant for Modern mode?

Thanks.

pbendersky:
@cotman 
- That is correct. Setting a parameter with a default value of `*/master` (Jenkins default) works as you expect: the web hook triggers builds normally for any branch with pushes / MR, and you can build manually and specify the branch to build (screenshots below).
- The setting called "Use GitLab CI features" is probably a misnomer. It actually means ["Use GitLab''s commit status API"](http://doc.gitlab.com/ce/api/commits.html#commit-status). When it's selected, the plugin will mark the builds in GitLab as passing or not. IMHO, having this granularity is useful only on Legacy mode, and on Modern mode it should be set by default. I left it as is since maybe there's some use case I'm missing where you want builds to trigger but the status not to be shown in GitLab.

screen shot 2016-04-01 at 9 59 49 am
screen shot 2016-04-01 at 9 59 58 am

cotman:
@pbendersky Excellent - thanks.  Look forward to this making it towards the released plugin.

pbendersky:
@omehegan @coder-hugo not sure who are the maintainers of this project, but saw you have several commits. Is there anything I can do to help getting this merged?

coder-hugo:
@pbendersky I hadn't yet the time for a deeper look to your changes. But your branch is based on the current master which will be replaced in a few days by the release-1.2 brach which contains a lot of code cleanups (almost everything was touched during the cleanup). Therefore I'd suggest you to have a look to the release-1.2 branch and adapt your changes accordingly. This will definitively help getting it merged.

pbendersky:
@coder-hugo I'll look into `release-1.2` and update this MR for it.

pbendersky:
@coder-hugo I see that you've released 1.2. I'll work into updating this MR for it.

pbendersky:
I just created #285 reimplementing this for 1.2 (BTW, the new architecture made it really easy to do, kudos). I'll close this PR, please keep the conversation in #285.

PR #258, opened 2016-04-05
Added gitlabBefore and gitlabAfter

pfrank13:
In the case of a Push Event the gitlabBefore and gitlabAfter parameters will be populated with the gitlab Push Event document's "before" and "after" attributes.  In the case of a merge request an empty string is added as it seems that is the project's default for StringParameterValue's that are absent


omehegan:
@pfrank13 can you rebase this change against the current master? Then we can include it in the next release. Thanks!

omorillo:
@pfrank13 Could you please update the **README** to include this new feature. Thanks!

pfrank13:
At this point this merge request is so far out of date with the current architecture that I should just redo the work to add it back in when I pierce the veil on where it would go in the current architecture.  Feel free to cancel this pull request and I'll resubmit when I get some time to look through the code again.

omehegan:
@pfrank13 OK, just today I had someone ask me for essentially this feature, so I would like to see it!

Issue #280, opened 2016-05-02
Tag Push Events not Starting Builds


kerlandsson:
We are trying to get Tag push events to work in GitLab Community Edition 8.7.0. Using the latest code from master of this plugin. When a tag is pushed, our job is not started. We have enabled FINE logging and see the following logs in Jenkins:

```
May 02, 2016 9:09:32 AM INFO com.dabsquared.gitlabjenkins.webhook.GitLabWebHook getDynamic
WebHook called with url: /project/test-pu-job
May 02, 2016 9:09:32 AM FINE com.dabsquared.gitlabjenkins.webhook.ActionResolver
Unsupported event header: Tag Push Hook
```

Looking at the code of `com.dabsquared.gitlabjenkins.webhook.ActionResolver.onPost(AbstractProject<?, ?>, StaplerRequest)` we see the following:

```
String eventHeader = request.getHeader("X-Gitlab-Event");
if (eventHeader.equals("Merge Request Hook")) {
    return new MergeRequestBuildAction(project, requestBody);
} else if (eventHeader.equals("Push Hook")) {
    return new PushBuildAction(project, requestBody);
}
LOGGER.log(Level.FINE, "Unsupported event header: {0}", eventHeader);
return new NoopAction();
```

I.e., when the header is `Tag Push Hook`, a `NoopAction` is returned. What is the intent - should a `PushBuildAction` be returned or is ? If so, I can submit a simple PR.

omehegan:
Version 1.2.3 of the plugin was just released and has this fix.

PR #281, opened 2016-05-02
Fixed issue where builds would not be triggered on tag push events

kerlandsson:
Fix for #280. I assume this is the way it is intended to work. It works in our setup at least.

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/webhook/ActionResolver.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/webhook/ActionResolver.java
@@ -95,7 +95,7 @@ public class ActionResolver {
         String eventHeader = request.getHeader("X-Gitlab-Event");
         if (StringUtils.equals(eventHeader, "Merge Request Hook") ) {
             return new MergeRequestBuildAction(project, requestBody);
-        } else if ( StringUtils.equals(eventHeader,"Push Hook")) {
+        } else if ( StringUtils.equals(eventHeader,"Push Hook") || StringUtils.equals(eventHeader,"Tag Push Hook")) {
             return new PushBuildAction(project, requestBody);
         }
         LOGGER.log(Level.FINE, "Unsupported event header: {0}", eventHeader);
--- a/src/test/java/com/dabsquared/gitlabjenkins/webhook/ActionResolverTest.java
+++ b/src/test/java/com/dabsquared/gitlabjenkins/webhook/ActionResolverTest.java
@@ -138,6 +138,21 @@ public class ActionResolverTest {
         assertThat(resolvedAction, instanceOf(PushBuildAction.class));
     }
 
+    @Test
+    public void postPushTag() throws IOException {
+        String projectName = "test";
+        jenkins.createFreeStyleProject(projectName);
+        when(request.getRestOfPath()).thenReturn("");
+        when(request.getMethod()).thenReturn("POST");
+        when(request.getHeader("X-Gitlab-Event")).thenReturn("Tag Push Hook");
+        when(request.getInputStream()).thenReturn(new ResourceServletInputStream("ActionResolverTest_postPushTag.json"));
+
+        WebHookAction resolvedAction = new ActionResolver().resolve(projectName, request);
+
+        assertThat(resolvedAction, instanceOf(PushBuildAction.class));
+    }
+
+
 
     private static class ResourceServletInputStream extends ServletInputStream {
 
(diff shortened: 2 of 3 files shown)


brunojdo:
I have this problem and I hope it will be accepted as soon as possible !!

andreascian:
I got the same problem.. I'm looking forward for the next gitlab-plugin release :+1:

omehegan:
@kerlandsson I merged this change and tested it locally, but it still does not trigger a build for me. When I just do
`git tag foo`
`git push --tags`

The Jenkins log now says:

May 13, 2016 10:00:57 PM INFO com.dabsquared.gitlabjenkins.trigger.handler.AbstractWebHookTriggerHandler handle
test-gitlab-hook triggered for push.

That's as opposed to the 'Unsupported event header' error that I was getting before. But nothing further happens after that. I have a feeling that more needs to be changed here but I'm not sure what.

omehegan:
@kerlandsson it seems like some other recent change in master is breaking the plugin for me, it's not your code. I'll work it out. Disregard my earlier comment (deleted).

omehegan:
OK, when I tested this on top of the patch which fixes the master branch, this change works as expected and I see no issues. I'm OK to merge it. @coder-hugo any concerns?

omorillo:
@kerlandsson Thanks for also providing some unit tests.

omorillo:
@kerlandsson @omehegan @coder-hugo 

I just had a quick look and to me the logic part looks good. For ease of _readability_ and _maintainability_ I would suggest a tiny change, namely the use of a switch case here:

``` java
private WebHookAction onPost(Job<?, ?> project, final StaplerRequest request) {
   final String requestBody = getRequestBody(request);
   final String eventHeader = request.getHeader("X-Gitlab-Event");
   switch(eventHeader) {
        case "Merge Request Hook":
             return new MergeRequestBuildAction(project, requestBody);
        case "Push Hook":
        case "Tag Push Hook":
             return new PushBuildAction(project, requestBody);
        default:
             LOGGER.log(Level.FINE, "Unsupported event header: {0}", eventHeader);
             return new NoopAction();
   }
}
```

What do you think?

kerlandsson:
@omorillo You'd need to add a null check on `eventHeader` before the switch statement in that case, or a NPE will be thrown if the header is not set. Something like:

``` java
private WebHookAction onPost(Job<?, ?> project, final StaplerRequest request) {
   final String requestBody = getRequestBody(request);
   final String eventHeader = request.getHeader("X-Gitlab-Event");
   if (evenHeader == null) {
        LOGGER.log(Level.FINE, "No event header set");
        return new NoopAction();
   }
   switch(eventHeader) {
        case "Merge Request Hook":
             return new MergeRequestBuildAction(project, requestBody);
        case "Push Hook":
        case "Tag Push Hook":
             return new PushBuildAction(project, requestBody);
        default:
             LOGGER.log(Level.FINE, "Unsupported event header: {0}", eventHeader);
             return new NoopAction();
   }
}
```

I'm not sure it became more readable after that, but we got the benefit of clearer logging in case the header is missing entirely. I'm fine with either approach.

omorillo:
@kerlandsson True. I didn't consider the implicit null check. It is truly not much more readable but still I would favour the switch solution for the explicit log output. Furthermore, we get rid of the _StringUtils_ import which is only used in this function. Take whatever you prefer. If you go for changing it, consider getting rid of the extra _requestBody_ variable to limit the scope of the _"getRequestBody"_ function call.

``` java
    private WebHookAction onPost(Job<?, ?> project, final StaplerRequest request) {
        final String eventHeader = request.getHeader("X-Gitlab-Event");

        if (eventHeader == null) {
            LOGGER.log(Level.FINE, "No event header set");
            return new NoopAction();
        }

        switch (eventHeader) {
            case "Merge Request Hook":
                return new MergeRequestBuildAction(project, getRequestBody(request));
            case "Push Hook":
            case "Tag Push Hook":
                return new PushBuildAction(project, getRequestBody(request));
            default:
                LOGGER.log(Level.FINE, "Unsupported event header: {0}", eventHeader);
                return new NoopAction();
        }
    }
```

kerlandsson:
I'd prefer not to change anything right now since @omehegan has tested the PR with the current contents, and the change as it stands now is really small. I can submit a PR for the refactoring after this one is merged.

omorillo:
Totally fine! :+1:

mcgrogan91:
Awesome.  Thanks for the quick fix for this.  Can't wait for the next version to get this fixed and put into use!

Issue #155, opened 2015-12-11
Error adding commits to existing Merge Request


pbendersky:
I'm experiencing an error when adding commits to an existing Merge Requests. The error I see, however, is different to the one described in #90, so I'm opening a new issue.

When creating a new MR, Jenkins build just fine. When adding a commit, however, I get the following:

```
Started by GitLab push by Pablo Bendersky
[EnvInject] - Loading node environment variables.
Building on master in workspace /Users/administrator/.jenkins/jobs/quadion.citest/workspace
 > git rev-parse --is-inside-work-tree # timeout=10
Fetching changes from 2 remote Git repositories
 > git config remote.pbendersky/citest.url git@gitlab.myserver.com:pbendersky/citest.git # timeout=10
Fetching upstream changes from git@gitlab.myserver.com:pbendersky/citest.git
 > git --version # timeout=10
 > git -c core.askpass=true fetch --tags --progress git@gitlab.myserver.com:pbendersky/citest.git +refs/heads/*:refs/remotes/pbendersky/citest/*
 > git config remote.origin.url git@gitlab.myserver.com:quadion/citest.git # timeout=10
Fetching upstream changes from git@gitlab.myserver.com:quadion/citest.git
 > git -c core.askpass=true fetch --tags --progress git@gitlab.myserver.com:quadion/citest.git +refs/heads/*:refs/remotes/origin/*
 > git rev-parse 71949a0e639735dabc40a84673fbed37db8632a8^{commit} # timeout=10
 > git branch -a --contains 71949a0e639735dabc40a84673fbed37db8632a8 # timeout=10
 > git rev-parse remotes/pbendersky/citest/more-tests^{commit} # timeout=10
Merging Revision 71949a0e639735dabc40a84673fbed37db8632a8 (pbendersky/citest/more-tests) to origin/more-tests, UserMergeOptions{mergeRemote='origin', mergeTarget='${gitlabTargetBranch}', mergeStrategy='default', fastForwardMode='--ff'}
 > git rev-parse origin/more-tests^{commit} # timeout=10
FATAL: Command "git rev-parse origin/more-tests^{commit}" returned status code 128:
stdout: origin/more-tests^{commit}

stderr: fatal: ambiguous argument 'origin/more-tests^{commit}': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'

[...]

Issue #98, opened 2015-08-05
Merge Request from Gitlab does not Merge but Checkout


AndreasBoehme:
When Jenkins receives a Merge-Request from Gitlab, it seems not to do the requested merge, but a checkout of a revision from the last merge-attempt (rev a15f43ab9677d5e2273a3d64d8e4e3be2e2d8ad5). That last merge-attempt failed too, and was meant to merge from a different branch.
So, I get why it is failing, but I don't get why it behaves that way.

This seems like a bug, or am I missing something here?

P.S.: Jenkins ver. 1.612, GitLab Plugin v1.1.25, git 2.5.0 on Ubuntu 12.04.2 LTS with Oracle JDK 1.8, GitLab 7.4.3

Log from the Build-Console:
(at the line it says 'Checking out Revision...' it should be something like 'Merging...' which I've seen in working Builds, but don't have the logs anymore)
```
GitLab Merge Request #10 : git-versioning => dev
Building in workspace /var/lib/jenkins/jobs/MyProject_Build/workspace
Wiping out workspace first.
Cloning the remote Git repository
Cloning repository git@git.mycompany.com:i2subst/MyProject.git
 > /usr/lib/git-core/git init /var/lib/jenkins/jobs/MyProject_Build/workspace # timeout=10
Fetching upstream changes from git@git.mycompany.com:i2subst/MyProject.git
 > /usr/lib/git-core/git --version # timeout=10
 > /usr/lib/git-core/git -c core.askpass=true fetch --tags --progress git@git.mycompany.com:i2subst/MyProject.git +refs/heads/*:refs/remotes/origin/*
 > /usr/lib/git-core/git config remote.origin.url git@git.mycompany.com:i2subst/MyProject.git # timeout=10
 > /usr/lib/git-core/git config --add remote.origin.fetch +refs/heads/*:refs/remotes/origin/* # timeout=10
 > /usr/lib/git-core/git config remote.origin.url git@git.mycompany.com:i2subst/MyProject.git # timeout=10
Pruning obsolete local branches
Fetching upstream changes from git@git.mycompany.com:i2subst/MyProject.git
 > /usr/lib/git-core/git -c core.askpass=true fetch --tags --progress git@git.mycompany.com:i2subst/MyProject.git +refs/heads/*:refs/remotes/origin/* --prune
Seen branch ...
...
Seen branch in repository origin/dev
Seen branch in repository origin/dev_INPD-374
[...]

PR #285, opened 2016-05-02
Adds a Modern Mode handling for merge requests.

pbendersky:
In Modern Mode, the plugin will fetch Merge Requests from origin instead of using parameterized builds.

This supersedes #252 and works with the refactored 1.2 codebase.


pbendersky:
Here's the original comment I wrote on #252:

This PR enhances gitlab-plugin by adding two operation modes: **Modern** and **Legacy**.
**Legacy** is what we currently have, that works with older versions of GitLab.

**Modern** works with GitLab 8.1 or better, and provides an easier configuration of projects in Jenkins (a single Git repository is enough) and much better handling of Merge Requests (MR are fetched from origin and matched by SHA1 instead of comparing branch names).

This PR fixes the two issues I reported: #239 and #155.

omehegan:
@pbendersky thanks a lot for doing this work!

src/main/java/com/dabsquared/gitlabjenkins/util/CommitStatusUpdater.java
```diff
@@ -5,6 +5,7 @@
 import com.dabsquared.gitlabjenkins.gitlab.api.GitLabApi;
 import com.dabsquared.gitlabjenkins.gitlab.api.model.BuildState;
 import hudson.EnvVars;
+import hudson.model.AbstractBuild;
```
omehegan: Are you using AbstractBuild here? Note that, in order to support Pipeline jobs, AbstractJob and AbstractBuild can't be used. See https://github.com/jenkinsci/pipeline-plugin/blob/master/DEVGUIDE.md

pbendersky: I'm not. I'll add a commit removing it.

pbendersky:
@coder-hugo @omehegan I'm keeping this PR updated with master, so it's easier to merge when you want to.
Please let me know if you want to do additional changes, or discuss any decisions I made.

omehegan:
@pbendersky thanks! I'm going to try to test it in my environment tomorrow, where we use GitLab CE 7.14. That's probably the most I can do with it. @coder-hugo said that he will look through it as soon as he can.

omehegan:
@pbendersky I'm working on testing this. The only merge conflict is with changes to the README, so it should be easy to resolve that.

pbendersky:
Nice! Let me know if you need anything, or you have comments on the code.

omehegan:
@pbendersky there seems to be a bug in the existing HEAD of master, I'm getting an NPE when I trigger a build with that code. Unrelated to your change. I'll see if I can track it down.

tommyminds:
Is there any update on the status of this PR? Or on the next release (1.2.3) in general? There are some fixes in master that would be good to have released.

omehegan:
I determined yesterday that the current master has a bug that breaks compatibility with GitLab 7.14 and possibly everything before GitLab 8.5. I have a patch for it that just needs cleanup. I will merge that this week and then I can test this change.

omehegan:
OK, I have successfully tested this with GitLab 7.14, using the legacy mode, and it correctly triggers a build based on push or MR, and checks out the right branch. So that much looks good to me.

src/main/java/com/dabsquared/gitlabjenkins/GitLabPluginMode.java
```diff
@@ -0,0 +1,32 @@
...
+package com.dabsquared.gitlabjenkins;
+
+/**
+ *
+ * @author Pablo Bendersky
+ */
+public enum GitLabPluginMode {
+    LEGACY,
+    MODERN;
+/*
```
omehegan: Do we need to keep all this commented out stuff?

src/main/java/com/dabsquared/gitlabjenkins/GitLabPluginMode.java
```diff
@@ -0,0 +1,32 @@
...
```
omehegan: Maybe remove this generic license boilerplate.

src/main/java/com/dabsquared/gitlabjenkins/cause/CauseData.java
```diff
@@ -152,7 +152,7 @@ public String getMergeRequestDescription() {
     public Integer getMergeRequestId() {
         return mergeRequestId;
     }
-
+
```
omehegan: Whitespace?

src/main/java/com/dabsquared/gitlabjenkins/trigger/handler/merge/AbstractMergeRequestHookTriggerHandler.java
```diff
@@ -0,0 +1,47 @@
+/*
```
omehegan: License boilerplate.

omehegan:
I posted some minor comments on the code. Also wanted to note that I tested with a Pipeline job and it works; this should imply that it will work with Freestyle jobs as well.

@coder-hugo pending comments from you before we merge this. I've reviewed as much as I think I can.

pbendersky:
@omehegan Just fixed these minor issues, and merged from upstream, to make things easier.

README.md
```diff
@@ -21,7 +21,11 @@
     - [Parameterized builds](#parameterized-builds)
 - [Branch filtering](#branch-filtering)
 - [Build Tags](#build-tags)
+<<<<<<< HEAD
```
omehegan: Looks like you committed the merge conflicts of README...

pbendersky: Sorry, just added a new commit fixing it.

README.md
```diff
@@ -39,10 +50,17 @@ Version 1.2.0 of the plugin introduced improved logging for debugging purposes.
...
+# Plugin Modes and Supported GitLab versions
+
+The Plugin supports two operation modes: **Modern** and **Legacy**. The main difference between these modes is how they deal with Merge Requests from GitLab. While the Legacy mode requires setting up multiple Git repositories and using parameterized builds, the Modern mode works with a single Git repository and relies on GitLab's availability of Merge Requests from the `origin` repository. The rest of this document is split for these two modes of operation (unless specified).
 
-* 7.14.x where it emulates Jenkins as a GitLabCI Web Service
-* 8.1.x and newer via the new commit status API that supports with external CI services like Jenkins
+The following versions of GitLab are supported by each of the modes:
+
+* Modern
+  * 8.1.x and newer
```
omorillo, line 70: Why is version **8.1.x** listed in both **Modern** and **Legacy**? The modes should be mutually exclusive, shouldn't they?

pbendersky, line 70: If you are using **>= 8.1** **Modern** is definitely better IMO. That said, nothing prevents you to use the **Legacy** mode. While I'd argue against these, I think there are some scenarios where you might want to leave **Legacy** mode:
1. Don't want to (or can't) mess with your Jenkins setup.
2. You want Jenkins to build pushes _on forks_ (**Modern** will build MR and commits added to MR, not commits to forks)

Basically, **Legacy** still works and **Modern** is opt-in (it's not even the default, although I think it has big advantages). If we want to push people to use **Modern**, we can change the wording here.

omehegan:
@coder-hugo ping

src/main/java/com/dabsquared/gitlabjenkins/trigger/handler/push/PushHookTriggerHandlerFactory.java
```diff
@@ -12,20 +13,20 @@
...
     }
 
-    private static List<PushHookTriggerHandler> retrieveHandlers(boolean triggerOnPush, TriggerOpenMergeRequest triggerOpenMergeRequestOnPush) {
+    private static List<PushHookTriggerHandler> retrieveHandlers(GitLabPluginMode gitLabPluginMode, boolean triggerOnPush, TriggerOpenMergeRequest triggerOpenMergeRequestOnPush) {
         List<PushHookTriggerHandler> result = new ArrayList<PushHookTriggerHandler>();
         if (triggerOnPush) {
             result.add(new PushHookTriggerHandlerImpl());
         }
-        if (triggerOpenMergeRequestOnPush == TriggerOpenMergeRequest.both) {
+        if (gitLabPluginMode == GitLabPluginMode.LEGACY && triggerOpenMergeRequestOnPush == TriggerOpenMergeRequest.both) {
```
coder-hugo, line 35: With this you disable the functionality to rebuild open MRs if the target branch has changed for the modern mode. Why you are doing this?

pbendersky, line 35: Because that case is handled by the Merge Request hook instead (if a MR is created and a commit added, the MR hook is called again).

coder-hugo, line 35: Nope. I just tested this with GitLab 8.7:
1. I configured the job in Jenkins as you described in the README
2. I configured my GitLab repo to send MR events to the Jenkins
3. I created a MR for a test branch into the master -> a new build was triggerd
4. I pushed some changes to the master (target branch) -> no build was triggered as no event was sent by GitLab

pbendersky, line 35: You are right. The case I was referring was new commits on the source branch (the user's fork). That gets handled by the MR trigger. I'll take a look at this, it might be a limitation of the "Modern" approach (if that's the case, I'll document it to reflect it).

pbendersky, line 35: @coder-hugo just looked a bit at this... seems to be a limitation of the Modern mode. Since the user only has one repo configured, a new commit to `master` in the target branch can't be merged to the MR by Jenkins.
I suggest:
1. Updating the Readme to reflect this.
2. Ideally (I'll check how can it be done) change the drop down when selecting Modern mode, so the user can only select "Source branch" on when to re-build.
Thoughts?

pbendersky:
@coder-hugo just added a two commits:
- I clarified in the Readme that in Modern mode only pushes to the target repo will trigger a rebuild.
- I added a `target` value to `TriggerOpenMergeRequest.java`
- I changed `GitLabPushTrigger.java` so it reloads the options for `TriggerOpenMergeRequest` when you change the plugin mode and only shows the valid values for each mode.
  Hope this makes it clearer to users.

I also merged the latest changes from master, so it passes the checks here.

coder-hugo:
@pbendersky it looks like you have interchanged the meaning of source and target branch of a MR. The source branch is the one that contains the changes that shall be merged to the target branch. So the modern mode supports re-building open MRs by pushes to the source but not to the target branch. And as the target branch is part of the configured repository it should be possible to re-build open MRs also for pushes to the target branch.

pbendersky:
Oh... I'll revert the source / target then.

So, a better approach (that would cover all the cases) would be to instead of fetching from `merge-requests/x` to fetch from `origin` and _merge_ the contents of `merge-requests/x`. That way, commits added to either source or target would get applied. Would that work? (of course, respecting the user settings for this).

pbendersky:
@coder-hugo I see one problem, however. Let's say there are 3 merge requests open, and a user commits to `master` in the upstream repository. How many builds should that trigger? If this works as you suggest, we would need to trigger four builds: one for the push to the upstream repo, and 3 for the updated merge requests. Is that correct?

omorillo:
@pbendersky I don't see any problem there. You would like to know if the master is still stable but also if the source branches can be safely merged to the target branch.

pbendersky:
@omorillo @coder-hugo I'm trying to wrap this up... Pending tasks (as discussed):
1. Merge origin with the merge request before build.
2. Add ability to trigger multiple builds on open merge requests when pushing to the target repo.

Now I need some help, and maybe you can give me a pointer... We currently create a `RevisionParameterAction` to tell Jenkins what revision to checkout. Is there any `Action` I can use to tell it to merge two repositories? I could not find it, and I'm not that familiar with Jenkins' API yet.

coder-hugo:
@pbendersky you don't need to add a special `Action` to the build for merging two branches. This will be configured as additional behavior (Merge before build) of the git SCM config. (see [here](https://github.com/jenkinsci/gitlab-plugin#jenkins-job-configuration))

pbendersky:
@coder-hugo right, but it loses some of the charm. As it is now, you simply set up `origin` on Jenkins and it will know how to fetch the merge requests from it. I'd like to keep that simplicity if possible. Maybe adding a second repository on build time? Other idea?

coder-hugo:
@pbendersky you can try to extend the `PreBuildMerge` of the git-plugin and pass the required `UserMergeOptions` for GitLab within the constructor to the super constructor. This extended class should then appear as a new additional behavior for the git SCM. This should keep the job setup as simple as possible.

pbendersky:
@coder-hugo @omehegan please take a look at #370. While I'm not giving up on this PR (I still think the setup with a single repository is easier to manage), the minor changes in configuration explained in #370 achieve the same result.

pabloa:
Hello, any news with this Pull Request? I am interested because #98

pbendersky:
I think it's a dead end at this point... With the configuration now in the README, I don't think this is needed. Did you try that?

omehegan:
Should we make a call on this one? It sounds like we should close it. Agreed?

pbendersky:
Agreed.

PR #289, opened 2016-05-03
Swap equals check around to prevent NPE

demaniak:
Swapped equals check on strings around to prevent NPE and let the code handle the missing header condition as intended.

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/webhook/ActionResolver.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/webhook/ActionResolver.java
@@ -90,9 +90,9 @@ public class ActionResolver {
     private WebHookAction onPost(Job<?, ?> project, StaplerRequest request) {
         String requestBody = getRequestBody(request);
         String eventHeader = request.getHeader("X-Gitlab-Event");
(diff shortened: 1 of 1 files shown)


demaniak:
This should prevent the NPE mentioned in #287 

It will of course not magically make the missing header appear :)

coder-hugo:
I don't like this type of comparing strings. As this changes the meaning of the "if-sentence". The current code says: I want to know if the value of the variable eventHeader is (equals) `Merge Request Hook` or `Push Hook`. Your code says: I want to know if `Merge Request Hook` or `Push Hook` is the value of the variable eventHeader.
Jenkins has already the apache commons-lang as a dependency so we should use the StringUtils for a null-safe and syntactical correct equals check.

coder-hugo:
Can you just squash your commits to a single one then I'll merge this PR back.

demaniak:
@coder-hugo oke-doke, I will try to get to that during today.

demaniak:
huh, @coder-hugo find myself unsure how to safely squash the commits (never done before, and I quickly settle this before going home)

I'm just going to close this PR, and re-submit with single commit.

PR #310, opened 2016-05-11
Fix getMergeRequests API

WonderCsabo:


coder-hugo:
@WonderCsabo can you just explain what this fix is for?

WonderCsabo:
Sorry for not doing that before. After updating to Gitlab 8.7, the plugin
failed because rest easy could not parse the response of this call. After
changing this, now it is working. This code also works with Gitlab 8.3.

coder-hugo:
But with this change you will not get all open merge requests as the response is always paged. Also GitLab 8.7 sends the pagination headers. Can you please provide the concrete error that happens with the current implementation.

WonderCsabo:
You are right. This is still paginated, even if i do not add the pagination parameters. I was mislead because the number of open MRs in our repo was less then the default per_page parameter value (20).

I just tested again, and i do not see this issue. Maybe we used a stale Gitlab version which was fixed later. :S

PR #311, opened 2016-05-11
Fix finding related commit of the build

WonderCsabo:
It seems we have to use the "marked" commit and not the "revision". For example when an MR is built, we merge the MR to the target branch first. In this case "marked" is the head of the source branch, and "revision" is the created merge commit on top of the target branch. But we want to update the status of the head of the source branch, which is "marked".

I found this issue when the plugin tried to update not existing commits, which turned out to be the merge commits created during the Jenkins build.


src/main/java/com/dabsquared/gitlabjenkins/util/CommitStatusUpdater.java
```diff
@@ -64,7 +72,15 @@ private static void printf(TaskListener listener, String message, Object... args
...
+        BuildData action = build.getAction(BuildData.class);
+        try {
+            Field lastBuildField = action.getClass().getDeclaredField("lastBuild");
+            lastBuildField.setAccessible(true);
+            Build lastBuild = (Build) lastBuildField.get(action);
+            return lastBuild.getMarked().getSha1String();
+        } catch (NoSuchFieldException | IllegalAccessException e) {
+            return null;
+        }
     }
```
WonderCsabo, line 70: I did not find a better way than this. 😢

WonderCsabo:
@coder-hugo i updated the test as well.

src/main/java/com/dabsquared/gitlabjenkins/util/CommitStatusUpdater.java
coder-hugo, line 70: You could try the method getLastBuild(ObjectId sha1) of the BuildData and pass the result of getLastBuildRevision().getSha1(). If this works we would get rid of the ugly reflection stuff.

WonderCsabo:
@coder-hugo good idea, this way it seems to be working. PR updated.

omehegan:
@WonderCsabo are you referring to using the 'Publish build status to GitLab commit' post-build step? Just checking, if so I can't test this because I'm still on GitLab 7.14.

WonderCsabo:
Yes.

src/main/java/com/dabsquared/gitlabjenkins/util/CommitStatusUpdater.java
MartinNowak, line 70: lastBuild is a public field of BuildData, and up to now there was no null checking for the result of `getLastBuiltRevision`, so it seems to be always set.
https://github.com/jenkinsci/git-plugin/blob/7ea2ad06da30aa6778b33fdd4c1f94389c20fbee/src/main/java/hudson/plugins/git/util/BuildData.java#L41

WonderCsabo, line 70: Wow, i am blind. Or i really got familiar with fields are never public in Java. :)

WonderCsabo, line 70: @coder-hugo should i change to use the public field, or the current method is okay (an indirection, but does the same thing)?

ipulyahin:
Hello guys!
I updated plugin to 1.2.3 version, but CI build status for merge requests(in GitLab) still remains skipped.
GitLab showed "0 builds for commit xyznasdf" although jenkins build succesfull triggered.
Environment:
- GitLab Community Edition 8.6.4 e47b581
- Jenkins ver. 1.651.2

PR #322, opened 2016-05-17
Git push hook namespace compatibility for Gitlab pre ~v8.5

xinranxiao:
Gitlab hooks pre ~v8.5 do not send the "project" field with the POST json. If we want to support using this plugin for versions of gitlab before ~8.5, we need try populating the project field with values that can be inferred from fields that _do_ exist in the json.

This patch simply fills in the namespace field in the project field for `PushBuildAction`'s pushHook if it does not get parsed in from the POST json. We need this because `pushHook.getProject().getNamespace()` is used later in the plugin code, and it will throw a null pointer exception if we do not have the project filled out.

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/webhook/build/MergeRequestBuildAction.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/webhook/build/MergeRequestBuildAction.java
@@ -17,7 +17,7 @@ import static com.dabsquared.gitlabjenkins.util.JsonUtil.toPrettyPrint;
 /**
  * @author Robin Müller
  */
-public class MergeRequestBuildAction implements WebHookAction {
+public class MergeRequestBuildAction extends BuildWebHookAction {
 
     private final static Logger LOGGER = Logger.getLogger(MergeRequestBuildAction.class.getName());
     private Job<?, ?> project;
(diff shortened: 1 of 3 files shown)


omehegan:
@xinranxiao thanks Xinran!
@coder-hugo we need to get this in before next release. Any concerns?

src/main/java/com/dabsquared/gitlabjenkins/webhook/build/MergeRequestBuildAction.java
```diff
@@ -29,7 +29,12 @@ public MergeRequestBuildAction(Job<?, ?> project, String json) {
         this.mergeRequestHook = JsonUtil.read(json, MergeRequestHook.class);
     }
 
-    public void execute(StaplerResponse response) {
+    /**
+    * Noop for merge request hooks.
+    */
+    void processForCompatibility() {}
```
coder-hugo: As this method isn't implemented for a merge request why does the MergeRequestBuildAction extend the BuildWebHookAction at all? This class still just needs to implement the WebHookAction interface. This will make the  BuildWebHookAction obsolete as you can move the processForCompatibilty stuff directly to the PushBuildAction.

xinranxiao: The reason I made `BuildWebHookAction` was because other build hook types could also need to use the compatibility steps.

For example, merge request webhooks have a bunch of fields that only exist in 8.x versions of gitlab and not the 7.x versions, such as source->homepage and source->url  (both of which we end up using in `MergeRequestHookTriggerHandlerImpl`). I added an implementation for this method for MergeRequestBuildAction in d9a66ae5ba77268b4324bddefd1dfdbb545a28c3.

omehegan:
Confirmed that this solves the problem we were having with 7.14. I'm using this to help test other PRs that have been submitted.

Issue #295, opened 2016-05-04
NPE during the build of a merge request


angely-dev:
Hi,

I'm trying to get Jenkins building the source branch merged with the target branch of a Merge Request. Hence, I followed [these instructions](https://github.com/jenkinsci/gitlab-plugin#jenkins-job-configuration), including the _Additional Behaviours_ step.

When a Merge Request is created in GitLab, Jenkins runs the build as expected but then a NullPointerException is triggered:

```
Merging Revision 73bf4ee77234767a99467a60af19d700db98ad94 (origin/feature/something) to origin/develop, UserMergeOptions{mergeRemote='origin', mergeTarget='${gitlabTargetBranch}', mergeStrategy='default', fastForwardMode='--ff'}
 > git rev-parse origin/develop^{commit} # timeout=10
 > git config core.sparsecheckout # timeout=10
 > git checkout -f origin/develop
 > git merge --ff 73bf4ee77234767a99467a60af19d700db98ad94 # timeout=10
 > git config core.sparsecheckout # timeout=10
 > git checkout -f 73bf4ee77234767a99467a60af19d700db98ad94
FATAL: null
java.lang.NullPointerException
    at hudson.plugins.git.extensions.impl.PreBuildMerge.decorateRevisionToBuild(PreBuildMerge.java:88)
    at hudson.plugins.git.GitSCM.determineRevisionToBuild(GitSCM.java:1009)
    at hudson.plugins.git.GitSCM.checkout(GitSCM.java:1098)
    at hudson.scm.SCM.checkout(SCM.java:485)
    at hudson.model.AbstractProject.checkout(AbstractProject.java:1269)
    at hudson.model.AbstractBuild$AbstractBuildExecution.defaultCheckout(AbstractBuild.java:607)
    at jenkins.scm.SCMCheckoutStrategy.checkout(SCMCheckoutStrategy.java:86)
    at hudson.model.AbstractBuild$AbstractBuildExecution.run(AbstractBuild.java:529)
    at hudson.model.Run.execute(Run.java:1738)
    at hudson.model.FreeStyleBuild.run(FreeStyleBuild.java:43)
    at hudson.model.ResourceController.execute(ResourceController.java:98)
    at hudson.model.Executor.run(Executor.java:410)
Finished: FAILURE
```

My set up:
![jenkins1](https://cloud.githubusercontent.com/assets/4362224/15016983/051c602a-1214-11e6-92a8-b99e93f6f417.PNG)
[...]

Issue #284, opened 2016-05-02
NullPointerException when using the GitLabCommitStatusPublisher


mattpfr:
Hi,
I am using the gitlab-plugin v.1.2.1 with Jenkins 1.646 and Gitlab 8.6.5 (CE).
I have a proper connection to gitlab and wanted to use the publisher functionality to get back the status from the job for a merge request.
If I do not add the publish post-build action, everything works fine, except that I do not get the build status in Gitlab.
If I now add the publisher post-buil action, I get an error when the build is executed: 

``` java
 > git rev-list 035fea1192d1533fcbf8cee8dc769c695df8ebeb # timeout=10
FATAL: null
java.lang.NullPointerException
    at org.jboss.resteasy.util.ReadFromStream.readFromStream(ReadFromStream.java:30)
    at org.jboss.resteasy.client.jaxrs.internal.ClientResponse.bufferEntity(ClientResponse.java:307)
    at org.jboss.resteasy.client.jaxrs.internal.proxy.extractors.DefaultEntityExtractorFactory$3.extractEntity(DefaultEntityExtractorFactory.java:48)
    at org.jboss.resteasy.client.jaxrs.internal.proxy.ClientInvoker.invoke(ClientInvoker.java:104)
    at org.jboss.resteasy.client.jaxrs.internal.proxy.ClientProxy.invoke(ClientProxy.java:64)
    at com.sun.proxy.$Proxy94.headCommit(Unknown Source)
    at com.dabsquared.gitlabjenkins.publisher.GitLabCommitStatusPublisher.existsCommit(GitLabCommitStatusPublisher.java:93)
    at com.dabsquared.gitlabjenkins.publisher.GitLabCommitStatusPublisher.updateCommitStatus(GitLabCommitStatusPublisher.java:78)
    at com.dabsquared.gitlabjenkins.publisher.GitLabCommitStatusPublisher.prebuild(GitLabCommitStatusPublisher.java:48)
    at hudson.model.AbstractBuild$AbstractBuildExecution.preBuild(AbstractBuild.java:837)
    at hudson.model.AbstractBuild$AbstractBuildExecution.preBuild(AbstractBuild.java:832)
    at hudson.model.Build$BuildExecution.doRun(Build.java:144)
    at hudson.model.AbstractBuild$AbstractBuildExecution.run(AbstractBuild.java:534)
    at hudson.model.Run.execute(Run.java:1738)
[...]

PR #323, opened 2016-05-18
Allign to GitLab API

bulanovk:
This PR fix 2 issues:
https://github.com/jenkinsci/gitlab-plugin/issues/295
https://github.com/jenkinsci/gitlab-plugin/issues/284

According to GitLab Commit API - http://docs.gitlab.com/ee/api/commits.html
all operations performed on commit objects are requires to provides ProjectId as integer instead of String used in com.dabsquared.gitlabjenkins.util.CommitStatusUpdater#retrieveGitlabProjectIds

To be alligned with GitLab API
com.dabsquared.gitlabjenkins.util.CommitStatusUpdater#retrieveGitlabProjectIds was updated to send get-single-project(http://docs.gitlab.com/ee/api/projects.html#get-single-project) request and use real ProjectId(Int) provided by GitLab in response

Tested with
GitLab 8.6.1
Jenkins 1.637

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/util/CommitStatusUpdater.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/util/CommitStatusUpdater.java
@@ -4,19 +4,21 @@ import com.dabsquared.gitlabjenkins.cause.GitLabWebHookCause;
 import com.dabsquared.gitlabjenkins.connection.GitLabConnectionProperty;
 import com.dabsquared.gitlabjenkins.gitlab.api.GitLabApi;
 import com.dabsquared.gitlabjenkins.gitlab.api.model.BuildState;
-import hudson.EnvVars;
-import hudson.model.Run;
-import hudson.model.TaskListener;
-import hudson.plugins.git.util.BuildData;
-import jenkins.model.Jenkins;
 
-import javax.ws.rs.NotFoundException;
-import javax.ws.rs.WebApplicationException;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+
+import javax.ws.rs.NotFoundException;
+import javax.ws.rs.WebApplicationException;
+
+import hudson.EnvVars;
+import hudson.model.Run;
+import hudson.model.TaskListener;
+import hudson.plugins.git.util.BuildData;
+import jenkins.model.Jenkins;
 
 /**
  * @author Robin Müller
@@ -98,7 +100,7 @@ public class CommitStatusUpdater {
         List<String> result = new ArrayList<>();
         for (String remoteUrl : build.getAction(BuildData.class).getRemoteUrls()) {
             try {
-                result.add(ProjectIdUtil.retrieveProjectId(environment.expand(remoteUrl)));
+                result.add(getClient(build).getProject(ProjectIdUtil.retrieveProjectId(environment.expand(remoteUrl))).getId().toString());
             } catch (ProjectIdUtil.ProjectIdResolutionException e) {
                 // nothing to do
             }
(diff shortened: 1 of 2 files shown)


tommyminds:
Since I had the same problem as you describe (the string project id to be used instead of the integer), I just applied this PR to my fork, rebuild the plugin and tried it out but this PR seems to break things even more.

```
May 18, 2016 12:12:29 PM FINEST com.dabsquared.gitlabjenkins.gitlab.GitLabClientBuilder
Call GitLab:
HTTP method: GET
URL: https://sbp.gitlab.schubergphilis.com/api/v3/projects/api%2Fmicroservice
Request headers: [
Accept = [application/json],
]
May 18, 2016 12:12:29 PM FINEST com.dabsquared.gitlabjenkins.gitlab.GitLabClientBuilder
Got response from GitLab:
URL: https://sbp.gitlab.schubergphilis.com/api/v3/projects/api%2Fmicroservice
Status: 404 Not Found
```

Notice it is not using the project id integer to retrieve the project

bulanovk:
Hello, TommyM which version of GitLab do you use?

Could you try to execute this command?
curl -k -X GET  -H "PRIVATE-TOKEN: PUTTOKEN HERE" https://sbp.gitlab.schubergphilis.com/api/v3/projects/api%2Fmicroservice

tommyminds:
```
GitLab 8.6.8
GitLab Shell 2.6.12
GitLab API v3
```

Digging into it a bit further it seems that the apache reverse proxy that we have in front of our gitlab is not properly handling requests with %2F in them... If I do the following it works:

```
curl -k -X GET -H "PRIVATE-TOKEN: PUTTOKEN HERE" https://sbp.gitlab.schubergphilis.com/api/v3/projects/1
```

But this fails:

```
curl -k -X GET -H "PRIVATE-TOKEN: PUTTOKEN HERE" https://sbp.gitlab.schubergphilis.com/api/v3/projects/api%2Fmicroservice
```

At first I thought it was the Gitlab API returning a 404, but it seems that the apache reverse proxy gateway is not even proxying the request with %2F in it to the gitlab app server.

tommyminds:
Ok, sorry for the confusion. I just found out that apache apparently has a default config called `AllowEncodedSlashes` that is set to off ;)

bulanovk:
No worries,i hope it will work once you correct apache settings

coder-hugo:
@bulanovk the commit status API of GitLab also accepts the namespace project combination as id.
Btw. the NPEs in the mentioned issues aren't caused by the fact that the plugin calls the GitLab API with the namespace project combination as id.

bulanovk:
@coder-hugo  Could you be so kind to provide a link to GitLab docs where it is mentioned that namespace project combination as id is acceptable for commits API?

As i could see  here http://docs.gitlab.com/ce/api/commits.html it is requires Integer.

Gordiychuk:
Why PR was close without merge? We found rest api violation and fix it. Why for version 1.1.32 was use Integer as project id, but now you use only project name?

coder-hugo:
As you can see here http://docs.gitlab.com/ce/api/projects.html#get-single-project you can retrieve a project by its id or the namespace project combination. The repository and its commits are just sub-resources of a project. As the GitLab API is kind of a REST API you should be (in this case you are) able to retrieve sub-resources always starting from the path of the top-level resource.
If this change would be necessary I'd merge it back. But it isn't. It just add an unnecessary call to GitLab. And especially it doesn't fix the mentioned issues.

omehegan:
The issues you said were fixed with this patch have already been fixed in HEAD of master. They just have not been released. So if you applied your patch to master and tested it, it would look like your patch solved the problem.

bulanovk:
@coder-hugo i take a look gitlab source code and agree with you, api should work not only with integer parameter

@omehegan nope, i got this issue on master head today at morning, after additional investigation i figure out right testcase, it fails when you have some special symbol as a part of id, dot for example.
It looks like rest client do not encodes this simbol while gitlab waiting for it. In this case my patch solve issue, but not in best way. 

I'l try to play with url encoding to provide a patch.

P.s. curl request with encoded dots works greate

bulanovk:
According to final investigation it was a little bit different issue related to Rails dynamic routing, i open new issue https://github.com/jenkinsci/gitlab-plugin/issues/326 and provide PR for it

PR #324, opened 2016-05-18
don't silently ignore commit status updates

MartinNowak:
- remove existsCommit request since changeBuildStatus will already return 404
- failing changeBuildStatus already logs an error message, a mismatch
  of the commit hash indicates a broken setup, so this shouldn't be silent

also see #321


MartinNowak:
This doesn't resolve #321, but at least we now get an error message in the logs.

Issue #272, opened 2016-04-26
Error while testing GitLab's Jenkins CI Integration with Jenkins' GitLab Plugin


radium226:
Hello, 

I'm trying to make GitLab 8.7 (using directly https://gitlab.com) and Jenkins 2.0 work together, but I got this error from Jenkins when I test (using the Test button of GitLab's Jenkins CI Service interface): 

```
java.lang.reflect.InvocationTargetException
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:498)
    at org.kohsuke.stapler.Function$InstanceFunction.invoke(Function.java:320)
    at org.kohsuke.stapler.Function.bindAndInvoke(Function.java:163)
    at org.kohsuke.stapler.MetaClass$11.dispatch(MetaClass.java:378)
    at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:746)
    at org.kohsuke.stapler.Stapler.invoke(Stapler.java:876)
    at org.kohsuke.stapler.MetaClass$11.dispatch(MetaClass.java:380)
    at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:746)
    at org.kohsuke.stapler.Stapler.invoke(Stapler.java:876)
    at org.kohsuke.stapler.Stapler.invoke(Stapler.java:649)
    at org.kohsuke.stapler.Stapler.service(Stapler.java:238)
    at javax.servlet.http.HttpServlet.service(HttpServlet.java:790)
    at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:812)
    at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1669)
    at hudson.util.PluginServletFilter$1.doFilter(PluginServletFilter.java:135)
    at hudson.plugins.greenballs.GreenBallFilter.doFilter(GreenBallFilter.java:59)
    at hudson.util.PluginServletFilter$1.doFilter(PluginServletFilter.java:132)
    at hudson.plugins.scm_sync_configuration.extensions.ScmSyncConfigurationFilter$1.call(ScmSyncConfigurationFilter.java:49)
[...]

PR #321, opened 2016-05-18
 CI build status for merge requests remains skipped since updating to 1.2.x

MartinNowak:
Something went wrong with the gitlab build status update. Since updating to version 1.2.x (from 1.1.x) builds in Gitlab 8.7.5 CE get marked as skipped.
I updated to a 1.2.3-SNAPSHOT build from master in order to get the fix for https://github.com/jenkinsci/gitlab-plugin/issues/284, and I had to manually add post-build hooks to report the status (b/c the gitlab config got lost during updating).
Still the status API is not working. Can I get more details logs somewhere to see whether the plugin actually called gitlab and what happened? The status calls seem to arrive at gitlab, but sth. in their content might be wrong (really sending skip as action?, commit hash after merge?).

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/util/CommitStatusUpdater.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/util/CommitStatusUpdater.java
@@ -34,7 +34,7 @@ public class CommitStatusUpdater {
                     GitLabApi client = getClient(build);
                     if (client == null) {
                         println(listener, "No GitLab connection configured");
-                    } else if (existsCommit(client, gitlabProjectId, commitHash)) {
+                    } else {
                         client.changeBuildStatus(gitlabProjectId, commitHash, state, getBuildBranch(build), "jenkins", buildUrl, null);
                     }
                 } catch (WebApplicationException e) {
@@ -65,16 +65,6 @@ public class CommitStatusUpdater {
 
     private static String getBuildRevision(Run<?, ?> build) {
         return build.getAction(BuildData.class).getLastBuiltRevision().getSha1String();
-    }
-
-    private static boolean existsCommit(GitLabApi client, String gitlabProjectId, String commitHash) {
-        try {
-            client.getCommit(gitlabProjectId, commitHash);
-            return true;
-        } catch (NotFoundException e) {
-            LOGGER.log(Level.FINE, String.format("Project (%s) and commit (%s) combination not found", gitlabProjectId, commitHash));
-            return false;
-        }
     }
 
     private static String getBuildBranch(Run<?, ?> build) {
--- a/src/main/java/com/dabsquared/gitlabjenkins/util/CommitStatusUpdater.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/util/CommitStatusUpdater.java
@@ -64,7 +64,7 @@ public class CommitStatusUpdater {
     }
 
     private static String getBuildRevision(Run<?, ?> build) {
-        return build.getAction(BuildData.class).getLastBuiltRevision().getSha1String();
+        return build.getAction(BuildData.class).lastBuild.marked.getSha1String();
     }
 
     private static String getBuildBranch(Run<?, ?> build) {
--- a/src/test/java/com/dabsquared/gitlabjenkins/publisher/GitLabCommitStatusPublisherTest.java
+++ b/src/test/java/com/dabsquared/gitlabjenkins/publisher/GitLabCommitStatusPublisherTest.java
@@ -17,6 +17,7 @@ import hudson.model.Result;
 import hudson.model.StreamBuildListener;
 import hudson.model.TaskListener;
 import hudson.plugins.git.Revision;
+import hudson.plugins.git.util.Build;
 import hudson.plugins.git.util.BuildData;
 import hudson.util.Secret;
 import jenkins.model.Jenkins;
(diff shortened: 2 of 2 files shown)


MartinNowak:
Other people report the same issue https://github.com/jenkinsci/gitlab-plugin/issues/272#issuecomment-216759224.

MartinNowak:
No in fact the statuses are not posted to gitlab, I confused this with some other gitlab internal ci_status GET. No more `/api/v3/projects/27/statuses` POSTs since the update. Any recommendation on how I can debug this?

MartinNowak:
Found it, creating a new logger for `com.dabsquared.gitlabjenkins` ALL revealed all the necessary information.
The cause for this problem is indeed what I suspected, the gitlab plugin reports the status for the commit after merging that into master. So it gets a 404 from gitlab when asking for that specific commit, and concludes `Project (xxx/yyy) and commit (59d4db...) combination not found`.

NITEMAN:
Same here.
Gitlab 8.7.5-ce.0
Jenkins 2.5
Plugin 1.2.2

MartinNowak:
I suspect this was introduced here https://github.com/jenkinsci/gitlab-plugin/commit/796998e593373eddbd71518324846a8274c637b5#diff-777e0dadd41db11dfedb64a44c31b08fL68, where the official GitlabAPI client was exchanged w/ a self-written jax-rs/RESTEasy client. Not sure why this was done.

MartinNowak:
Not sure how this worked before, maybe it's just that the existsCommit step fails?
But looking at the other side, gitlab really requires a sha1 for an existing commit .
[lib/api/commit_statuses.rb](https://gitlab.com/gitlab-org/gitlab-ce/blob/da8ac1635f1c12e9455e7517bee29678c754d492/lib/api/commit_statuses.rb#L51)
Maybe something in jenkins' git-plugin or pipeline changed, that switched the revision associated with a build from pre-merge to post-merge.
At worst we could check for `run.getAction(MergeBuild.class).mergeRevision`, at best we could checkout gitlab's commit from the auto-merge instead of merging in jenkins.

I'll try to remove the `existsCommit` check first, it's an unnecessary request anyhow b/c the commit status API already deals with non-existing commits.

tommyminds:
For me the problem seems to be related to the plugin using a string as the project id instead of the integer. In the jenkins logs I see it make a call to the following url to get the commit from the gitlab API:

```
May 18, 2016 12:25:06 PM FINEST com.dabsquared.gitlabjenkins.gitlab.GitLabClientBuilder
Call GitLab:
HTTP method: GET
URL: https://sbp.gitlab.schubergphilis.com/api/v3/projects/api%2Fmicroservice/repository/commits/6c2673f2936f26f830391fb41cf28bebee0006a5
Request headers: [
Accept = [application/json],
]
May 18, 2016 12:25:06 PM FINEST com.dabsquared.gitlabjenkins.gitlab.GitLabClientBuilder
Got response from GitLab:
URL: https://sbp.gitlab.schubergphilis.com/api/v3/projects/api%2Fmicroservice/repository/commits/6c2673f2936f26f830391fb41cf28bebee0006a5
Status: 404 Not Found
```

If I use curl to change this to the following then the call works as expected:

```
https://sbp.gitlab.schubergphilis.com/api/v3/projects/1/repository/commits/6c2673f2936f26f830391fb41cf28bebee0006a5
```

MartinNowak:
Well I get a commit not found response in both cases, maybe there are two causes for this bug.
I can also run a GET request with either `/api/v3/projects/id` or `/api/v3/projects/name`, so I don't think this is the problem.
Do you merge before building?

tommyminds:
Yeah sorry for the confusion! Turns out the apache reverse proxy gateway we have in front of our Gitlab was blocking the requests with %2F in it because AllowEncodedSlashes is turned off by default in Apache.... So ignore my post, it is unrelated to this problem!

MartinNowak:
Converted this issue into a pull request.
See [git-plugin/Build.java#L41](https://github.com/jenkinsci/git-plugin/blob/ce5479a42e74b45dd8b954776ad4220c9c0151bc/src/main/java/hudson/plugins/git/util/Build.java#L41) for documentation of Build.marked, it will report the status of merge commits to the pre-merge commit.

coder-hugo:
This issue get already fixed by #311. The exists commit check is necessary if there are multiple repositories configured and the commit only exists in one of them. This shouldn't be treated as an error.
@MartinNowak btw. I replaced the "official GitLab java client" by a JAX-RS interface with resteasy as JAX-RS provider as this client is a real bad implementation. E.g. the whole URL writing is done by hand. There were several bugs in this client just because parts of the URL weren't encoded correctly. With JAX-RS you don't have such issues. Furthermore the "official client" doesn't provide the possibility to configure things like connection pooling and timeouts. And using JAX-RS provides a lot of other advantages. For example the client logging that prints the whole request and response of calls to GitLab to the log can be archived by just implementing a CleintRequestFilter and a ClientRepsoneFilter. In the "offical client" the only way to archive this would be doing it by hand for each and every method of the API.

MartinNowak:
> Furthermore the "official client" doesn't provide the possibility to configure things like connection pooling and timeouts. And using JAX-RS provides a lot of other advantages.

It wasn't an official client after all, just got misleaded by the org.gitlab name.

MartinNowak:
> The exists commit check is necessary if there are multiple repositories configured and the commit only exists in one of them. This shouldn't be treated as an error.

I see, handling NotFoundException when reporting the status would fix that and still save a request.

Issue #326, opened 2016-05-19
Support NameSpace or ProjectName with dot in it


bulanovk:
Support NameSpace or ProjectName with dot in it
According to Rails documentation(http://guides.rubyonrails.org/v3.1.3/routing.html section 3.2 Dynamic Segments) Rail use a dot as separator for formatted routes.
In this case if we have dot in ProjectId constructed from(namespace + projectname) all request except getSingleProject(http://docs.gitlab.com/ce/api/projects.html#get-single-project) fail with message
`Failed to update Gitlab commit status for project Name.Test/Repo': HTTP 500 Internal Server Error`

omehegan:
Released in version 1.2.3 of the plugin.

PR #327, opened 2016-05-19
Fix to support NameSpace or ProjectName with dot in it

bulanovk:
According to Rails documentation(http://guides.rubyonrails.org/v3.1.3/routing.html section 3.2 Dynamic Segments) Rail use a dot as separator for formatted routes.
In this case if we have dot in ProjectId constructed from(namespace + projectname) all request except getSingleProject(http://docs.gitlab.com/ce/api/projects.html#get-single-project) fail with message
"Failed to update Gitlab commit status for project Name.Test/Repo': HTTP 500 Internal Server Error"

To support This case we check if projectId have dot inside of it
if it has execute getSingleProject request and extract Integer projectId to operate with
if it hasn't use prepared projectId(Namespace + ProjectName)

Fix Issue https://github.com/jenkinsci/gitlab-plugin/issues/326

### DIFF
--- a/src/test/java/com/dabsquared/gitlabjenkins/publisher/GitLabCommitStatusPublisherTest.java
+++ b/src/test/java/com/dabsquared/gitlabjenkins/publisher/GitLabCommitStatusPublisherTest.java
@@ -21,6 +21,8 @@ import hudson.plugins.git.util.Build;
 import hudson.plugins.git.util.BuildData;
 import hudson.util.Secret;
 import jenkins.model.Jenkins;
+
+import org.apache.commons.io.IOUtils;
 import org.eclipse.jgit.lib.ObjectId;
 import org.hamcrest.CoreMatchers;
 import org.jenkinsci.plugins.plaincredentials.impl.StringCredentialsImpl;
@@ -34,6 +36,7 @@ import org.mockito.stubbing.Answer;
 import org.mockserver.client.server.MockServerClient;
 import org.mockserver.junit.MockServerRule;
 import org.mockserver.model.HttpRequest;
+import org.mockserver.model.HttpResponse;
 import org.mockserver.verify.VerificationTimes;
 
 import java.io.ByteArrayOutputStream;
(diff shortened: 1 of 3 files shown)


src/test/java/com/dabsquared/gitlabjenkins/publisher/GitLabCommitStatusPublisherTest.java
```diff
@@ -219,6 +236,62 @@ private HttpRequest prepareExistsCommit(String projectId, String sha) throws Uns
                 .withHeader("PRIVATE-TOKEN", "secret");
     }
 
+    private HttpRequest prepareGetProjectResponse(String projectName, int projectId) throws UnsupportedEncodingException {
+      HttpRequest request= request()
```
coder-hugo: Wrong indent.

src/test/java/com/dabsquared/gitlabjenkins/publisher/GitLabCommitStatusPublisherTest.java
```diff
@@ -219,6 +236,62 @@ private HttpRequest prepareExistsCommit(String projectId, String sha) throws Uns
...
+               .withPath("/gitlab/api/v3/projects/" + URLEncoder.encode(projectName, "UTF-8"))
+               .withMethod("GET")
+               .withHeader("PRIVATE-TOKEN", "secret");
+
+        HttpResponse response = response().withBody(
+            "{\"id\":"+projectId+","+
```
coder-hugo: Please add a space before and after each plus sign. Or even better move the JSON to a separate file with placeholders like in the `MergeRequestBuildActionTest`.

coder-hugo:
This ugly workaround wouldn't be necessary if rails would use HTTP as it was designed. IMHO rails is a very cool framework but using the URL for formatting the content differently really sucks. HTTP provides exactly for this purpose the Accept header.

bulanovk:
@coder-hugo yes, this is really strange behavior from Rails side, but in this case we have only 2 options to cover this behavior:
1) Introduced by this PR, get Integer projectId in case when we have dot's
2) Perform custom PathParams encoding and replace "." to "%2E", but it's more ugly than additional call to GitLab from my point of view

coder-hugo:
@bulanovk you are right. The second option could cause issues for other API calles where you don't expect it.
Can you just squash your 2 commits to a single one then I'll merge it back.

bulanovk:
Squash to one commit, rebase to master

PR #349, opened 2016-06-09
Fix NPE when updating commit status

WonderCsabo:
If there was no commit for the last build because the pushed commit no longer exists, a NPE could occur.


omorillo:
@WonderCsabo Thanks for your pr. Will have a look and merge if everything is alright.

omorillo:
Makes sense. If the checks complete successfully I will merge.

WonderCsabo:
Thanks for the quick review.

Issue #407, opened 2016-09-08
Failed to update the commit status with CI-Pipelines


Alien2150:
### Context
- **Gitlab plugin version**: 1.4.0
- **Gitlab version**: GitLab Community Edition 8.10.6
- **Jenkins version**: 2.2.1
### Logs & Traces

Logs: https://gist.github.com/Alien2150/af096e99cdb09eb7f39e89ab24968961
### Problem description

I setup a pipeline snippet (https://gist.github.com/Alien2150/4b300102fae862e86fce08281c7f6b38) and as soon as it reaches the "gitlabCommitStatus" it fails with the exception "Failed to update Gitlab commit status for project '135': java.lang.RuntimeException: com.fasterxml.jackson.core.JsonParseException: Illegal character ((CTRL-CHAR, code 31)): only regular white space (\r, \n, \t) is allowed between tokens" (See logs). Let me know If I forget to attach missing information. 

Regards,
Thomas

qrsty:
Try restarting your jenkins instance. I had this issue, and it went away after reboot.

Alien2150:
Thanks will try

Alien2150:
Still did not help. Did you restarted the whole server or only the service (Like /etc/init.d/jenkins restart)?

qrsty:
I used `http://jenkinsUrl.com/safeRestart`.

I am also using Jenkins 2.19, GitLab CE 8.11.2 with GitLab plugin 1.3.2 currently.

Alien2150:
So did:
- Rebooted the machine
- Restarted service with /etc/init.d
- Used safeRestart - url 

Did you put your build-scripts into scm? Might that be an issue or maybe permissions for the Gitlab User?

qrsty:
I was getting the json parsing error with any usage of the plugin, be it a FreeStyle job or pipeline. Can you check if a post-build action on a FreeStyle job works?

Alien2150:
Good point. Will try

Alien2150:
With the Freestyle-Job I got: 

is allowed between tokens
 at [Source: ????n?0%???$[?=??e?#??,fi?/?E;e??T??S?A?o?Z??p??^??!???_?e???+?r?d????p?G?E?g?<]??e??r???R?4 ????L#f?P?%?VB??#?e@?=?{?'?:k:k+x??????????????????4Dp?!????????s{r??z??%??[C??;_~>~?_?&Y?Js?n???R?; line: 1, column: 2]

[...]

PR #423, opened 2016-09-30
Added Remove-Accept Encoding Filter

Alien2150:
Fix for https://github.com/jenkinsci/gitlab-plugin/issues/407

For unknown reasons the HTTP Requests sometimes sets the Accept-Encoding to "gzip; deflate" which will lead to exceptions in the Jackson-Library. The remove-filter simply eliminates the "Accept-Encoding" Value


omehegan:
@coder-hugo thoughts on this patch? Looks pretty simple. This is impacting some others including @cliles.

PR #431, opened 2016-10-13
Removed dependency to git-plugin

HeikoBoettger:
Hi,
this is a branch derived from the branch for the pull request https://github.com/jenkinsci/gitlab-plugin/pull/401. Since I couldn't reach the owner, I create an new pull request.

The corresponding GitLab Support Ticket is 36768.

What I have done:
- Rebased the branch on the original pull request to the master
- Fixed compilation errors after rebasing caused by non-conflicting merges
- Removed all remaining dependencies to the jenkins git-plugin and removed it from pom.xml
- Updated the tests

Open points:
- I prepared the possibility to add a new option for only build the head of a branch but not a checkbox on the ui (commit-id should than be added to the same scheduled build as Cause)
- I didn't checked whether I have broken the serialization of the old settings
- I never installed the compiled version in a real environment
- I never checked whether the branch name is stored in the correct format. I saw that the formats used in the Merge Request Hook and Push Hook are different.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -158,6 +158,11 @@
     <!-- Jenkins dependencies -->
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>git</artifactId>
+      <version>2.4.1</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>git-client</artifactId>
       <version>1.19.0</version>
     </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -158,11 +158,6 @@
     <!-- Jenkins dependencies -->
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
-      <artifactId>git</artifactId>
-      <version>2.4.1</version>
-    </dependency>
-    <dependency>
-      <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>git-client</artifactId>
       <version>1.19.0</version>
     </dependency>
@@ -276,6 +271,11 @@
       <artifactId>postgresql</artifactId>
       <version>9.4.1208</version>
       <scope>test</scope>
+    </dependency>
+    <dependency>
+    	<groupId>org.jenkins-ci.plugins</groupId>
+    	<artifactId>scm-api</artifactId>
+    	<version>1.0</version>
     </dependency>
   </dependencies>
 
--- a/src/main/java/com/dabsquared/gitlabjenkins/GitLabPushTrigger.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/GitLabPushTrigger.java
@@ -79,6 +79,7 @@ public class GitLabPushTrigger extends Trigger<Job<?, ?>> {
     private boolean ciSkip = true;
     private boolean skipWorkInProgressMergeRequest;
     private boolean setBuildDescription = true;
+    private boolean alwaysBuildHead = false;
     private transient boolean addNoteOnMergeRequest;
     private transient boolean addCiMessage;
     private transient boolean addVoteOnMergeRequest;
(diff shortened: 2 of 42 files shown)


dblessing:
@omehegan Do you mind taking a look at this, please?

omehegan:
Can someone explain the reasoning behind this change? Why do we want to remove this dependency?

dblessing:
@omehegan Might be good to read https://github.com/jenkinsci/gitlab-plugin/pull/401 for context.

HeikoBoettger:
@omehegan 
The main reason are:
- a build doesn't strictly require to checkout any code in the triggered job (we use multiple downstream jobs)
- it's currently not possible to build a different branch or repository (parent project of a submodule) since the commitid from git-plugin is used in many places to detect scheduled build and notification.
- if using more than one entry in the git section to checkout code, always the first one is used
- probably rarely happening, but building for a merge request by rebase/merge the source-branch on to target-branch for building doesn't handle the case of multiple merge request with different target-branches (the plugin only uses the source-branch commit id)

Result:
- triggering jobs on every merge request change due to the missing lastcommit id
- non-working build event webhook (documentation?)

HeikoBoettger:
Hi,
can someone tell me where to find information about who the ui and serialization works and where I can find testcases which I need to take care. I really want to get this contribution mainline.

Please tell me if I need to take care about further things. It would be also nice if someone could tell me what's exactly the problem in the build. Is it only findbugs?

Best Regards
Heiko

dblessing:
@coder-hugo If you have a moment, will you please look at this and see what you think? Thanks for your help.

HeikoBoettger:
@benjie332 
Hi Benjie,
I saw you comment on the other pull request, that's right. My pull request basically contains all the changes. 

Have you had the time to look into this?

May be you have a hint, what needs to be done to turn the jenkins build from yellow to green.

omehegan:
@trance1st this PR has been open for some time, there is interest in getting it merged but I did not feel capable of giving it a serious review. Maybe you can do so when you have some time!

trance1st:
Hi all,
I am totally against this PR
I have two big concerns: 
- the merge have some failed tests (this can be solved) :  GitLabCommitStatusPublisherTest
- this PR breaks a main functionality of this plugin which is "publishing the build status to gitlab"

The second problem occurs when you have a simple Jenkins job with a post build action to publish result on GitLab. If you **manually** trigger the build then there is no hook from which we can extract infos so we have a problem.
This is the message from the Jenkins's job execution:
```
Failed to update Gitlab commit status: no associated GitLabWebHookCause
Finished: SUCCESS
```
and the reason is the class CommitStatusUpdater.

Another concern (a small one) is that there are same mixed features here: I understand and I like  the main reason to remove the dependency of the git-plugin but we can work together to find a viable solution, solve it first, and then add new features. For example this feature "only build the head of a branch" is totally a separate one.

Anyway, the last paragraph it not so important (it's my opinion to add small changes gradually to the code)

The important thing that we need to discuss here is the fact that with these changes the plugin will only work for MR.

HeikoBoettger:
@trance1st 
Hi, 
just for my understanding, what's the reason to notify GitLab on a manual build? Is it just the missing rebuild-button on the GitLab side?
In my environment we use one jenkins-job for all projects, so manually triggering won't work.

For me it's not a problem to add the git-plugin dependency again and restore the current behavour of the GitLab-Plugin "publishing the build status to gitlab"-functionality as a fallback. However the GitLab-Plugin should prefer to use it's own cause if it's present.

benjie332:
@HeikoBoettger  There was an earlier version of this pull request which did something like that - if it detected no Gitlab plugin presence it fell-back to using Jenkins data.

HeikoBoettger:
@benjie332 
Yes, I know. I need to restore the fall-back logic. I have a look at this.

HeikoBoettger:
Hi,
I have a small question regarding CommitStatusPngAction. Why does this use getBuildBySHA1WithoutMergeBuilds insteadof getBuildBySHA1IncludingMergeBuilds?
```
public CommitStatusPngAction(Job<?, ?> project, String sha1) {
        super(project, BuildUtil.getBuildBySHA1WithoutMergeBuilds(project, sha1));
}
```
I am asking because it's seem strange for me that the StatusJsonAction isn't using the same method. Isn't that resulting into potential inconsistencies between the icon and status?

HeikoBoettger:
@benjie332 
I restored the fallbacks and updated the related tests. However although the test are running on my machine. 

@trance1st 
Can someone please explain me the following build error? Is this some random maven build error?

[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] Unresolveable build extension: Plugin org.jenkins-ci.tools:maven-hpi-plugin:1.115 or one of its dependencies could not be resolved: Failed to collect dependencies at org.jenkins-ci.tools:maven-hpi-plugin:jar:1.115 -> org.kohsuke.stapler:stapler-groovy:jar:1.237 -> org.kohsuke.stapler:stapler-jelly:jar:1.237 -> org.jenkins-ci:commons-jelly:jar:1.1-jenkins-20120928 @ 
[ERROR] Unknown packaging: hpi @ line 20, column 14

omehegan:
@HeikoBoettger are the tests running OK for you locally? If so this could have been a transient problem with the jenkins.io build farm.

omehegan:
Reopen if there is still interest in working on this.

Issue #446, opened 2016-11-15
Environment variables not expanded for build name in "Publish build status to Gitlab commit"


kbthomp1:
### Context
- **Gitlab plugin version**: 1.4.3
- **Gitlab version**: 8.13.1
- **Jenkins version**: 2.30

### Problem description
The **Publish build status to Gitlab commit** post-build feature doesn't seem to be expanding environment variables correctly.  In this case, I'm triggering a jenkins build using a java regex, and then parsing the string to get the Jenkins build name to trigger.  I then use the [EnvInject](https://wiki.jenkins-ci.org/display/JENKINS/EnvInject+Plugin) plugin to set the environment variable "TestTriggerName".  This method works for the **Add note with build status on Gitlab merge request** post-build feature, where I define a custom message for success/fail/abort that includes the TestTriggerName variable.  for the **Add note with build status on Gitlab merge request** feature, ${TestTriggerName} is correctly expanded and posted to gitlab; however in the **Publish build status to Gitlab commit** feature, the build name doesn't expand correctly. Listing the variable for the build name in jenkins:

![jenkins_side](https://cloud.githubusercontent.com/assets/16530166/20307826/45779ec0-ab0f-11e6-82f7-a642a37f453b.png)

produces the variable as literal on the gitlab side:

![bad_expansion_example](https://cloud.githubusercontent.com/assets/16530166/20307748/df1a9d4e-ab0e-11e6-8831-14549506d0de.png)

I would have expected the TestTriggerName variable to be expanded as the build note was in this case.

omehegan:
This is a plugin bug. We will have to explicitly expand this text field in the plugin code. Reference [this answer.](http://stackoverflow.com/questions/23800416/jenkins-inject-environment-variable-in-jelly-field)

trance1st:
Hi all
@omehegan @kbthomp1
Please see my comments here https://github.com/jenkinsci/gitlab-plugin/pull/479

We can add this feature but it will work only for manually triggered jobs (like the use case described).

In case of MR builds we have a problem: the plugin first receives the hook from gitlab, it does same logic (to reject or accept the request etc) and publish the build to gitlab with status = "pending" and then it schedules the job. At this point, right before the build actually starts, we cannot expand variables because we don't have a running context of the job.

@omehegan 
Do you agree this bug to be solved only for manually triggered builds?
We can add info to the documentation to explain that the 'build name' field can contain variables only for manually triggered jobs.
[...]

PR #479, opened 2017-01-08
[WIP] Issue #446 : expand build name

trance1st:
Fix for #446

Adapted GitLabCommitStatusPublisher to expand the 'build name' field.

Still need to do some work because of those hooks that updates the build status to 'pending'. 
The problem is that those classes **are executed before the build is scheduled** and we don't have access to the build context/environment to expand a variable (see below)

See publisher.getName() invocation from AbstractWebHookTriggerHandler
`client.changeBuildStatus(buildStatusUpdate.getProjectId(), buildStatusUpdate.getSha(), BuildState.pending, buildStatusUpdate.getRef(),
                                             **publisher.getName()**, Jenkins.getInstance().getRootUrl() + job.getUrl() + job.getNextBuildNumber(), null);`

Inside AbstractWebHookTriggerHandler we need a way to expand variables. As is decribed here (http://stackoverflow.com/questions/23800416/jenkins-inject-environment-variable-in-jelly-field), we need the Run and BuildListener objects to expand a variable.
 
The 'Run' class can be obtained from this Future:
`private void scheduleBuild(Job<?, ?> job, Action[] actions)`
I can add a callback here but then, the 'pending' status will be displayed for a very short period of time because the 'GitLabCommitStatusPublisher.prebuild' method will be called which updates the status to 'running'.

**Shortly**: the 'pending' status on builds it's useful if the build stays too long in the Jenkins queue. We can't add the feature to expand some variables at this point because there is no instances of the job and if we will wait for the build to start running then the status is logically 'running'.

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/publisher/GitLabCommitStatusPublisher.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/publisher/GitLabCommitStatusPublisher.java
@@ -8,6 +8,7 @@ import hudson.model.AbstractBuild;
 import hudson.model.AbstractProject;
 import hudson.model.BuildListener;
 import hudson.model.Result;
+import hudson.model.Run;
 import hudson.tasks.BuildStepDescriptor;
 import hudson.tasks.BuildStepMonitor;
 import hudson.tasks.Notifier;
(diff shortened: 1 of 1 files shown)


omehegan:
@trance1st is this PR still a WIP or is it done now?

omehegan:
@trance1st ping.

omehegan:
Reopen if you want to get this merged.

Issue #452, opened 2016-11-23
Extension for "Add message only for failed builds"


sauliusgrigaitis:
Thanks a lot for resolving #345 !

We started to use Unstable status for builds that passes tests and doesn't pass linters. This is very convenient way to indicate what is wrong with not passing build. So now we need to add message to Gitlab for both Unstable and Failed builds (#345 allows to add message only for Failed builds). 

So would be great to have one of these options:
1) Checkbox "Add message only for failed *and unstable* builds" instead of current "Add message only for failed builds"
2) Two checkboxes "Add message only for failed builds" and "Add message only for unstable builds". Labels need some work, because if both are checked, then "only" words doesn't make sense.

omehegan:
It looks like you're working on a fix, we would be happy to merge it or help with the design if you need it.

sauliusgrigaitis:
Thanks @omehegan, feel free to merge this fix until somebody implements full configurable feature.

omehegan:
@sauliusgrigaitis can you submit it as a PR?

omehegan:
This was released today in version 1.4.8 of the plugin.

PR #483, opened 2017-01-09
Send GitLab messages on unstable builds as well as failed ones

sauliusgrigaitis:
I'm submitting this small fix for #452. It's not full configurable feature as described in #452, but it does the job without allowing to configure details.


omehegan:
I think we should expand on this to make it configurable before we merge it.

sauliusgrigaitis:
Well, question is does anybody really need that flexible configuration. I personally think that those people who use Unstable status in their builds would also like to get notifications just like they get for Failed builds. Those who doesn't use Unstable status at all will even not notice this merge request.

omehegan:
@sauliusgrigaitis sorry for the long delay on this. I wanted to merge this today, but tests are failing. Can you try rebasing with master, and let's see if the tests pass now?

sauliusgrigaitis:
@omehegan done!

omehegan:

PR #503, opened 2017-02-13
Append a "/" to the build URL for "pending" state

### LINKED ISSUE #489: "Pending" commit status is never updated, instead new status is created
### Context

- **Gitlab plugin version**: 1.4.0
- **Gitlab version**: 8.15.4-ee
- **Jenkins version**: 1.650

### Logs & Traces

target_url when creating "pending" status:
```json
"target_url" : "https://server:443/jenkins/job/GitlabMergeRequests/25"
```
target_url when creating/updating "running" status:
```json
"target_url" : "https://server:443/jenkins/job/GitlabMergeRequests/25/"
```

### Problem description

We have a jenkins job setup to build all merge requests (triggered via a gitlab webhook with the GitLab CI Service URL). The job also has the post-build steps **Publish build status to gitLab commit** and **Add note with build status on GitLab merge requests**. This is working very well, except that for every build a "pending" commit status is left behind. The "pending" status is created at the start of the build, but once the build starts a new commit status with status "running". That commit status is then updated with the result of the build once it finishes.

I looked through the logs and think the problem is that the `target_url` has a trailing slash when creating/updating the second status, but when the initial "pending" status is created there is no such trailing slash.

hho:
The URL generated for the "pending" state was different from the URL of the finished build.

Closes #489

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/trigger/handler/AbstractWebHookTriggerHandler.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/trigger/handler/AbstractWebHookTriggerHandler.java
@@ -69,8 +69,10 @@ public abstract class AbstractWebHookTriggerHandler<H extends WebHook> implement
                 if (client == null) {
                     LOGGER.log(Level.SEVERE, "No GitLab connection configured");
                 } else {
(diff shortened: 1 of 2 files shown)


omehegan:
Thanks!

omehegan:
Version 1.4.5 was released just now with this fix, it should be in the update center within the next few hours.

Issue #512, opened 2017-02-22
Merge request CI is repeated when pushing to branch with same suffix as targetBranch name


czoellin:
## Issue
```
if (targetBranch != null && branchFilter.isBranchAllowed(targetBranch) && hook.getRef().endsWith(targetBranch) && sourceBranch != null) {
```

Seems to compare with `endsWith` probably to avoid having to deal with 'origin/' prefix?
Have not actually tried to fix this.

### Problem description

Basically title says it all.
We have several branches that have the suffix "_master" as well as an actual master branch.
Merge requests that target "master" are repeated, whenever anyone pushes to one of the "*_master" branches. This is a large repo with many contributors and a relatively length merge request CI of 30 minutes.

When there's a lot of activity this causes our merge requests to never successfully complete a jenkins job.

omehegan:
Interesting. Your theory about the cause seems plausible. Perhaps it would be better to use a regex here to strip off origin/. Would you like to attempt a fix?

czoellin:
After some more investigation, the comparison is actually against the `ref` field of the push object.
The value of that will usually be a git ref, like `refs/heads/master` for pushes to branches.
I think the proper solution is not to match anything, but testing for equality with `targetBranch` prepended by `refs/heads/`. `refs/heads/` is the only prefix that makes sense, because we do not care about tags that are pushed or other refs.

I will give the fix a try and test it out on our server.

omehegan:
This was released today in version 1.4.8 of the plugin.

PR #514, opened 2017-02-24
Fix branch name comparison to avoid spurious builds

czoellin:
Fixes issue #512.


omehegan:
Thanks! We'll consider this for the next release.

omehegan:
@czoellin is it possible this could break merge request builds, because they may check out refs/merge-requests/*/head ? I'm not totally clear on how this works. But see https://docs.gitlab.com/ee/user/project/merge_requests/index.html

omehegan:
@czoellin ping re: my last comment.

czoellin:
The target branch name that is being compared here is never going to be one of these hidden refs for merge requests. Don't see a problem.

omehegan:
OK, thanks! Will consider this for the next release.

omehegan:

PR #532, opened 2017-03-25
Adds custom Unstable textbox 

DiogoLG:
We can add a custom message for FAILED, SUCCESS or ABORTED builds.

This MR adds the option to add a custom message for UNSTABLE builds as well. 

Does not change the current behavior to consider a build as success or failed (or +1/-1 a MR).


DiogoLG:
@omehegan could you take a look at this?

omehegan:
Yep, as soon as I have some time :) Hopefully this week.

src/main/java/com/dabsquared/gitlabjenkins/publisher/GitLabMessagePublisher.java
```diff
@@ -75,6 +83,10 @@ public String getAbortNoteText() {
         return this.abortNoteText == null ? "" : this.abortNoteText;
     }
 
+    public String getUnstableNoteText() {
+        return this.unstableNoteText == null ? "" : this.unstableNoteText;
```
arielb2, line 87: When ternary is used and we have binary operators it is recommended to use parenthesis.
Ex.
<code>
   return (this.unstableNoteText == null) ? "" : this.unstableNoteText;
</code>

DiogoLG: I will update that, but I just follow the pattern that was in the code already

omehegan:
This was released today in version 1.4.6 of the plugin.

PR #508, opened 2017-02-17
Trigger build when merge request has been merged or closed

m9abbas:
Hello, the main of this pull request is to be able to trigger builds when a merge request has been accepted or closed (Please see #390 ).

PR #542, opened 2017-04-25
add Getter for new fields triggerOnAcceptedMergeRequest and triggerOnClosedMergeRequest

m9abbas:
The main of this PR is to fix the problem reported by @mreichel in PR #508. It adds 2 `Getter`s for fields `triggerOnAcceptedMergeRequest` and `triggerOnClosedMergeRequest`

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/GitLabPushTrigger.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/GitLabPushTrigger.java
@@ -193,6 +193,14 @@ public class GitLabPushTrigger extends Trigger<Job<?, ?>> {
         return triggerOnMergeRequest;
     }
 
+    public boolean isTriggerOnAcceptedMergeRequest() {
+        return triggerOnAcceptedMergeRequest;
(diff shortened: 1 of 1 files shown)


colourmeamused:
Thanks, looks good and we should merge @omehegan @mreichel any comments?

mreichel (approved):
I tested the PR in my Jenkins and the bug is fixed. If @omehegan agrees we can merge

omehegan:
👍

mreichel:
@omehegan @colourmeamused since I merge the PR, the dependency CI keeps on failing for the lib org.apache.httpcomponents:httpclient. There are no changes in the pom/dependency. Do you have any idea how I could fix this?

m9abbas:
I found an opened issue about this problem [here](https://github.com/dependencyci/support/issues/53).

m9abbas:
@mreichel @omehegan @colourmeamused, the problem seems to be resolved, please see [here](https://github.com/dependencyci/support/issues/53#issuecomment-299325023)

PR #543, opened 2017-04-27
Matrix/Multi-config support

evgeni:
This PR provides basic Matrix Job support by:
* injecting the environment variables also into MatrixRuns (the axis jobs)
* making the publishers MatrixAggregatable and thus allowing Flexible Publish and similar to execute them *after* all axis jobs have finished

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,19 @@ pipeline {
 ```
 
 ### Matrix/Multi-configuration jobs
-**The Jenkins Matrix/Multi-configuration job type is not supported.**
+
+This plugin can be used on Matrix/Multi-configuration jobs together with the [Flexible Publish](https://plugins.jenkins.io/flexible-publish) plugin which allows to run publishers after all axis jobs are done.
+
+To use GitLab with Flexible Publish, configure the *Post-build Actions* as follows:
+
+1. Add a *Flexible publish* action
+2. In the *Flexible publish* section:
+      1. *Add conditional action*
+      2. In the *Conditional action* section:
+          1. Set *Run?* to *Never*
+          2. Select *Condition for Matrix Aggregation*
+          3. Set *Run on Parent?* to *Always*
+          4. Add GitLab actions as required
 
 ## Gitlab Configuration
 
--- a/pom.xml
+++ b/pom.xml
@@ -192,6 +192,11 @@
       <artifactId>org.eclipse.jgit</artifactId>
       <version>3.5.2.201411120430-r</version>
     </dependency>
+    <dependency>
+        <groupId>org.jenkins-ci.plugins</groupId>
+        <artifactId>matrix-project</artifactId>
+        <version>1.10</version>
+    </dependency>
 
     <!-- REST client dependencies -->
     <dependency>
--- a/src/main/java/com/dabsquared/gitlabjenkins/environment/GitLabEnvironmentContributor.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/environment/GitLabEnvironmentContributor.java
@@ -3,6 +3,8 @@ package com.dabsquared.gitlabjenkins.environment;
 import com.dabsquared.gitlabjenkins.cause.GitLabWebHookCause;
 import hudson.EnvVars;
 import hudson.Extension;
+import hudson.matrix.MatrixRun;
+import hudson.matrix.MatrixBuild;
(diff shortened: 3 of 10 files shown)


omehegan:
Thanks very much for this PR! This has long been an issue for a minority of our users, so I appreciate you doing the work to get it supported! We'll try to get this reviewed soon.

evgeni:
Yak shaving at it's finest for me :-)

There are two obvious "issues" with the PR:
* I did not write any tests, mostly because my Java is rusty and I did not really understand how the mocking is done
* It is still not possible to submit GitLab data from an axis job (it would need similar handling like the EnvironmentContributor), but I did not care yet as we don't need that functionality

omehegan:
@evgeni when you say it is not possible to submit GitLab data, do you mean sending the build result back to GitLab? Or something else?

evgeni:
Assume you have the following matrix job (in jjb syntax):

```yaml
name: test-python
project-type: matrix
axes:
    - axis:
        type: python
        values:
        - 2.7
        - 3.5
```

If you set a "normal" post-build action in Jenkins, it happens at the end of each axis job `test-python/python=2.7` and `test-python/python=3.5`. Due to the fact that these are `MatrixRun`s, `parent.getCause(GitLabWebHookCause.class)` returns `null` and no submission of build results to GitLab can happen. No post-build action happens for `test-python` as that is a `MatrixBuild` and these for some reason don't trigger post-build actions (probably because they don't really build anything).

When using Flexible Publish, one can also assign a post-build action to the main `test-python` job by enabling the Matrix Aggregation. Here now `parent.getCause(GitLabWebHookCause.class)` returns the right config and the submission can happen. And for that I had to teach the publishers a bit to aggregate.

Thus, with this PR, you can submit `test-python` status to GitLab, but you cannot submit the individual axis statuses thus missing the detail which part exactly failed in the GitLab comment.

But I don't think this is too bad for a start? :)

omehegan:
@evgeni OK, that makes sense, and this lines up with my experience of using Matrix jobs (I had a love/hate relationship with them, and threw all mine away when Pipeline came along, but I understand there are still some things that Matrix does which Pipeline doesn't). I agree that this is a good start :)

So if I understand you correctly, if we merge this PR a user can build Matrix jobs from a GitLab trigger, and if they also use Flexible Publish they can add a post-build action which will report the status of the entire build (all axes of the matrix) to GitLab. And I see you've updated the README accordingly, thanks for that!

@mreichel or @colourmeamused could you possibly give @evgeni some guidance about adding tests for this feature? It's significant enough that I think that is important to have them before we merge it.

mreichel:
Most tests require a running Jenkins instance. You can use the `JenkinsRule` to launch an instance with the GitLab plugin running.

`@Rule public JenkinsRule jr = new JenkinsRule();`

In your tests, you can than access the variable to create and run projects and suchlike. 

Alternatively you can use mocks, if you don't need the full Jenkins instance (but I think for your pr the full instance is necessary).

As a rough idea: to test your pr, you could use JenkinsRule and create a matrix project (see [the matrix project's tests](https://github.com/jenkinsci/matrix-project-plugin/blob/master/src/test/java/hudson/matrix/MatrixProjectTest.java)
as an example on how to do it) and then try to trigger a build (similar to the existing tests in the GitLab project).
I don't know if that is helpful for you, @evgeni, otherwise just ping me.

evgeni:
Actually, I realized that we don't need full rules here and mocks are enough. We only have to make sure that the aggregation works and the publisher properly gets called. That is easily done in a mock. The rest of the code remains the same and works the same, no need to check twice, right?

evgeni:
~~So the only untested part is the `GitLabEnvironmentContributor`, but it had no tests at all before, and I am not able to make it work even without Matrix :/~~

evgeni:
Now with tests for all parts I touched. @omehegan @mreichel the stage is yours :D

omehegan:
@evgeni thanks very much!

mreichel (approved):
Looks good to me. Maybe some else can review as well, and afterwards we merge?

ehelms:
Looks good to me! Would love to get a release and this change into our environment.

ehelms (approved):
Would really appreciate a merge and release with this patch to unblock our CI.

omehegan:
This was released today in version 1.4.6 of the plugin.

Issue #427, opened 2016-10-05
Plugin should have secure defaults


omehegan:
Now that we have implemented a couple of methods for authenticating access for the GitLab webhooks, it would be a best practice to enable that authentication by default on new installs. This is in keeping with the Jenkins 2 philosophy of being secure by default.

My understanding is that we have two options for authentication. One is a global option where all access to the /project endpoint must use HTTP basic auth with a username and password that already have authorization to trigger jobs. The other is a per-job/per-project option where the user can specify a token on the GitLab webhooks page and the Jenkins job configuration page.

I would suggest that we enable the global security option by default. In concert with this, we should create a "Getting Started" section of the documentation (as part of refactoring all of the doc into the wiki) which explains how to use auth in the webhooks, as well as how to disable the auth if desired. I can handle the documentation work.

omehegan:
@daniel-beck when you have time, can you comment briefly on how I could implement a backward-compatible "enable security by default" setting? We talked about this once but I don't remember the details. The point being that I would only want to enable security on new installs of the plugin, not force-enable it when people upgrade. Thanks!

daniel-beck:
@omehegan 

> We talked about this once but I don't remember the details. The point being that I would only want to enable security on new installs of the plugin, not force-enable it when people upgrade. Thanks!

Sorry, drowned in GH notifications.

Have a `Boolean` (uppercase, i.e. nullable) field that's `true` by default, but gets set to `false` in `readResolve` if it's `null` (or the other way around). Databind a form field to it, and done.

The reason this works is that an instance of your persisted descriptor (or whatever you have) gets loaded on startup, going through `readResolve` rather than constructors/field initialization. It's `null` by default, so you know you loaded a pre-existing config. OTOH if someone sets up the configuration from scratch, the field default is used.

omehegan:
@mreichel or @colourmeamused would either of you want to make a pass at implementing this? It's been an unfortunate bug for quite awhile now that the plugin by default will allow unauthorized users to trigger builds. We want to change the default settings to be secure, but make sure we don't break anyone who is currently relying on the insecure configuration when they upgrade.

mreichel:
@omehegan, I can try. Just let me get this right: the "Enable authentication for '/project' end-point" functionality for the GitLab plugin settings in the global settings should be enabled by default, but only if it is not `null`. The authentication method  doesn't need to be changed.

omehegan:
@mreichel that's right. Valid tests:

[...]

PR #567, opened 2017-06-20
Plugin should have secure defaults

mreichel:
#427 Plugin should have secure defaults

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnectionConfig.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnectionConfig.java
@@ -35,7 +35,7 @@ import java.util.Map;
 @Extension
 public class GitLabConnectionConfig extends GlobalConfiguration {
 
-    private boolean useAuthenticatedEndpoint;
+    private Boolean useAuthenticatedEndpoint = true;
     private List<GitLabConnection> connections = new ArrayList<>();
     private transient Map<String, GitLabConnection> connectionMap = new HashMap<>();
     private transient Map<String, GitLabApi> clients = new HashMap<>();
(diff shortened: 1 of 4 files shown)


omehegan:
I just want to test this manually to confirm it's good before we merge.

src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnectionConfig.java
```diff
@@ -55,11 +55,11 @@ public boolean configure(StaplerRequest req, JSONObject json) throws FormExcepti
         return super.configure(req, json);
     }
 
-    public boolean isUseAuthenticatedEndpoint() {
+    public Boolean getUseAuthenticatedEndpoint() {
```
christ66: I am 👎 as this causes binary incompatibility. This should be `isUseAuthenticatedEndpoint` as this is a normal java convention for boolean properties`is<propertyName>(){....}`

christ66:
Other than the minor change, 👍

mreichel:
I changed the name, because otherwise it did not trigger.

christ66:
It might not work because you're using `Boolean` instead of `boolean` which it does not recognize the object version of boolean. I would recommend switching back to `boolean` as using the `Boolean` adds no value and it is possible to invoke `setUseAuthenticatedEndpoint(null)` which should not be possible.

mreichel:
Using the wrapper type is the idea of the commit because we need the null value (in [#427](https://github.com/jenkinsci/gitlab-plugin/issues/427) @daniel-beck is giving details on the original idea)

daniel-beck:
Right, but the API does not need changing; the API can remain `boolean`, you just need to make sure to have no paths that keep the `null` value between constructor and `readResolve` (and if you do it right, you won't).

mreichel:
Ok, I made the changes we discussed

daniel-beck:
This looks correct; but with something like this I recommend performing various upgrade tests.

omehegan:
Yes, I'm intending to set aside time to do some manual upgrade testing to make sure we don't mess anything up.

christ66 (approved):
Agree with the testing. Other than that it LGTM.

omehegan:
OK, I have tested this in the Docker setup that is included with the plugin. My process was:
1. Install 1.4.0 of the plugin and configure it to talk to GitLab. It defaults to no auth for the plugin endpoint.
2. Installed the current version of the plugin with this patch. Auth is still not enabled.
3. Uninstalled the plugin and manually cleared its config from JENKINS_HOME.
4. Installed the current version with patch again. Now auth shows up as enabled by default.

I pronounce this ready to merge. Hope we don't break anything!

daniel-beck:
🎉

omehegan:

PR #582, opened 2017-07-06
Change icon for build failure and aborted

chinhodado:
Right now we're using [this](https://emojipedia.org/negative-squared-cross-mark/) icon for build failure comment which in a lot of browsers is a green square, which is quite misleading at a quick glance. I changed it to [this](https://emojipedia.org/cross-mark/) icon instead.

Also the current icon for an aborted build is a [thumb up](https://emojipedia.org/white-up-pointing-index/), which doesn't make any sense, however I can't find any appropriate icon for "aborted" so I just removed it.

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/publisher/GitLabMessagePublisher.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/publisher/GitLabMessagePublisher.java
@@ -122,11 +122,11 @@ public class GitLabMessagePublisher extends MergeRequestNotifier {
         if (result == Result.SUCCESS) {
             return ":white_check_mark:";
         } else if (result == Result.ABORTED) {
-            return ":point_up:";
+            return "";
         } else if (result == Result.UNSTABLE) {
             return ":warning:";
         } else {
-            return ":negative_squared_cross_mark:";
+            return ":x:";
         }
     }
 


src/main/java/com/dabsquared/gitlabjenkins/publisher/GitLabMessagePublisher.java
```diff
@@ -122,11 +122,11 @@ private String getResultIcon(Result result) {
         if (result == Result.SUCCESS) {
             return ":white_check_mark:";
         } else if (result == Result.ABORTED) {
-            return ":point_up:";
+            return "";
```
hho, line 125: Maybe we could use the [octagonal sign](http://emojipedia.org/octagonal-sign/)? That would be `:octagonal_sign:` here.

Only downside is that neither Slack nor Mattermost support it yet.

chinhodado: I can only see a square on Chrome on Windows 8. Looks like this is not widely supported yet as it's part of Unicode 9.0 in 2016.

jaylinski: Maybe a warning icon ⚠️  would also be appropriate: http://emojipedia.org/warning-sign/

omehegan:
@chinhodado if you want to continue with this, let's get some consensus about what the best symbol would be. There seems to be some debate. The failing test build on ci.jenkins.io is probably transient so you can ignore that for now.

src/main/java/com/dabsquared/gitlabjenkins/publisher/GitLabMessagePublisher.java
hho: That's already used for "Unstable"...

hho: How about [Raised Hand :hand:](https://emojipedia.org/raised-hand/) `:hand:`? It's in Unicode 6.0 from 2010 and also available in Mattermost and Slack.

chinhodado: I think we can reuse the warning icon. The hand icon doesn't seem to fit the context.

hho: What's the point of having an icon in the first place, when different statuses map to the same icon?

chinhodado: The icon can signify the severity, not just individual status. Ideally we would have different icons for different statuses but if that's not possible because nothing fits "aborted" and is widely supported by all apps and browsers then it's not a bad compromise to let the icon to signify the severity instead.

hho: AFAICT a concept of "severity" does not exist in Jenkins. I also think it's questionable wheter a build which has completed (albeit unstable) should be the same "severity" as a build which hasn't completed at all (because it was aborted).

> nothing fits "aborted"

That is highly subjective. Even the Emojipedia page states that :hand: can mean "Stop"…

Trying to be constructive, though, here's another suggestion: :no_entry_sign: `:no_entry_sign:`

omehegan: This is an interesting problem, seems like we need a UX person to comment! ;)

✋  can mean stop but it also kind of looks like a high five to me.

In Jenkins itself, colored balls differentiate the build result. Blue for success, red for failure, yellow for "unstable" (used by some plugins and other custom cases, site-dependent), and grey for aborted.

Issue #580, opened 2017-07-06
Add separate gitlabActionType for a tag push event


bbreijer:
### Context
- **Gitlab plugin version**: 1.4.5
- **Gitlab version**: 9.1.0
- **Jenkins version**: 2.68
- **Job type**: Pipeline, but not it holds for all job types

### Problem description

On both `"object_kind": "push"` and `"object_kind": "tag_push"` the `gitlabActionType` is set to `PUSH`. This makes people abuse the `gitlabBefore` variable to determine a tag push while the correct information is available in the payload of the webhook.
Proposed solution: add a `TAG_PUSH` actionType.

omehegan:
I sort of see where this is happening in the code (https://github.com/jenkinsci/gitlab-plugin/blob/master/src/main/java/com/dabsquared/gitlabjenkins/cause/CauseData.java#L375) but I can't work out how exactly this influences the value of gitlabActionType. If anyone else wants to take this on, they can.

bbreijer:
I think the information is lost in the ActionResolver. [https://github.com/jenkinsci/gitlab-plugin/blob/master/src/main/java/com/dabsquared/gitlabjenkins/webhook/ActionResolver.java](url).
In `WebHookAction()`, both "Push Hook" and "Tag Push Hook" are handled the same. A distinction there could be the solution.
Another (imo better) solution is to change the point where the actual ActionType is set; in `PushHookTriggerHandlerImpl` which is now the same for both object kinds.

Changing the line `.withActionType(CauseData.ActionType.PUSH)` in `PushHookTriggerHandlerImpl` to something like `.withActionType(hook.getObjectKind().equals("tag_push") ? CauseData.ActionType.TAG_PUSH : CauseData.ActionType.PUSH)` and adding the `TAG_PUSH` type to the enum would be a possible solution.

I can change te code if this helps you, but i don't have any testing abilities. Would you like me to create an (untested) pull-request?

omehegan:
@bbreijer I see what you mean. Yeah if you could at least submit a PR, that would be helpful! It's a small enough change that some manual testing would probably be OK, or maybe someone else can add a unit test.

omehegan:
This has just been released in plugin version 1.5.2. It will be in the update center within a few hours.

gempain:
[...]

PR #589, opened 2017-07-17
Add support for tag_push event from GitLab

bbreijer:
#580 

### Before submitting a pull request, please make sure you read the instructions in the ["Contribution to the Plugin"](https://github.com/jenkinsci/gitlab-plugin/tree/master#contributing-to-the-plugin) section in the README.

*(if you read the above instructions, remove the paragraph and start describing the pull request)*


omehegan:
@bbreijer can you look at the test failures on this? Perhaps it need a rebase with master.

bbreijer:
@omehegan The tests succeed now, the object kind was missing in the hook builder.

omehegan:
@bbreijer thanks! From what I can tell of the diff, support for tag_push events is added, but there isn't any additional configuration option for it in the UI. So will tags always trigger builds if they are configured to trigger on pushes? It may be that users will want to disable that, and only trigger on code pushes, because that has been the behavior of the plugin until now.

bbreijer:
@omehegan The current behaviour is that the build is triggered on both a tag push and a regular push (if you configured the webhook in gitlab in that way). The only way to obtain the real cause of the build with this configuration is by looking at the branch value, this contains a tag if it was caused by a tag_push.
The changes i implemented make it more easy to distinguish between a regular push and a tag push.

I totally agree that a future improvement would be to add a setting with a tag specifier on which tags to build and not to build, but i think that the setting wether a build should trigger at all is already handled by the GitLab webhook configuration.

omehegan:
@bbreijer ok thanks, now I understand and this makes sense. Can you just clarify what someone would do in their build to make use of this? It's not apparent to me from the diff what they would do to check the cause.

bbreijer:
@omehegan One can use the environment variable `gitlabActionType` to distinguish between a `PUSH` and a `TAG_PUSH`, while this variable is set to `PUSH` in both situations now. The exact use is dependant on your personal situation of course.

omehegan:
@bbreijer aha, thank you, that was the part I was missing. Makes sense! I will merge this, but I am intending to do a release that only contains #614 first, once that is ready. Then I'll get this released.

cliles:
@omehegan - Any updates? TAG_PUSH would be great.

omehegan:

PR #524, opened 2017-03-14
If BlueOcean is installed, build URL does not point to BlueOcean

i386:
Fixes #518.
@omehegan @jkdeveyra this should fix the issue with pointing to Blue Ocean as mentioned in #518

Issue #518, opened 2017-03-03
If BlueOcean is installed, build URL does not point to BlueOcean


jkdeveyra:
### Context
- **Gitlab plugin version**: 1.4.5
- **Gitlab version**: 8.4.3
- **Jenkins version**: 2.45
- **Job type**:  Multibranch Pipeline
- **BlueOcean version**: 1.0.0-b24	

### Problem description

When I visit the build link in Gitlab:
![screen shot 2017-03-03 at 2 46 28 pm](https://cloud.githubusercontent.com/assets/570725/23541117/5238e392-0020-11e7-83f6-f3cc0bf5f345.png)

it points to the classic Jenkins build URL:
![screen shot 2017-03-03 at 2 48 45 pm](https://cloud.githubusercontent.com/assets/570725/23541227/ea8eec4a-0020-11e7-8817-a2557b472265.png)

instead of opening the link in BlueOcean:
![screen shot 2017-03-03 at 2 50 47 pm](https://cloud.githubusercontent.com/assets/570725/23541252/0ca3c224-0021-11e7-93a7-da033e503945.png)

omehegan:
Well, I can see why. The plugin tells GitLab what the URL should be to refer back to the build. The Blue Ocean URL is going to be different. And given that a user can switch in and out of the Blue Ocean view at will, I don't see a great way to change this. Just detecting that you have the plugin installed does not necessarily mean that you are using it in all cases. It's still in beta because not all necessary features are implemented in it yet. Also, different users of the same Jenkins install may prefer to use the old view or Blue Ocean.

@i386 this is an interesting situation. Do you have any thoughts on how we should approach it, now or in the future?

i386:
@omehegan I filed a PR for this but I am not sure if I found all the right places. Should be enough to go on if you want to continue :)

omehegan:
This has just been released in plugin version 1.5.2. It will be in the update center within a few hours.

Eldadc:
> This has just been released in plugin version 1.5.2. It will be in the update center within a few hours.

Hi,
Is it possible to control which Jenkins view can be opened ? I want my classic to be opened and Blue-ocean view is the now my default view.
Thanks In Advanced,

mjaggard:
@Eldadc Did you ever find a way to link to the better interface from your SCM?

Eldadc:
I am sending the URL of the classic web interface via email notifications.

PR #591, opened 2017-07-18
If BlueOcean is installed, build URL does not point to BlueOcean

ohartl:
Doing the changes at @i386 was about to do in PR #524. Since he is not continuing that PR soon, this replaces his PR.

@omehegan @jkdeveyra this should fix the issue with pointing to Blue Ocean as mentioned in #518

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -188,9 +188,14 @@
       <version>3.5.2.201411120430-r</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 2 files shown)


omehegan:
@ohartl did you change your mind about this?

ohartl:
Sorry i did. Since i couldn't figure things out with the project setup (not a java dev)...

PR #595, opened 2017-07-20
Recursive search of GitLabWebHookCause

Simmay93:
search recursively through upstream causes to find GitLabWebHookCause or return null.

With this change you can start an init your job from GitLab and e. g. notify GitLab when all downstream projects finished.

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/publisher/MergeRequestNotifier.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/publisher/MergeRequestNotifier.java
@@ -7,6 +7,8 @@ import hudson.matrix.MatrixAggregatable;
 import hudson.matrix.MatrixAggregator;
 import hudson.matrix.MatrixBuild;
 import hudson.model.AbstractBuild;
+import hudson.model.Cause;
+import hudson.model.Cause.UpstreamCause;
 import hudson.model.BuildListener;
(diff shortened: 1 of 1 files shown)


omehegan:
I'm not sure that I understand what this change accomplishes. If GitLab triggers a Jenkins job, and that job has downstream jobs, then I would have thought that the triggered job would not notify GitLab until all the downstream jobs are done.

Simmay93:
I thought that too. But actually the notification is done if the build is finished.

I tried to extend the plugin to notify when downstream projects are ready but it was too hard for me.

So currently I use my change as a work around to notify GitLab in the downstream job

omehegan:
@Simmay93 are these downstream jobs triggered as build steps, or post-build steps? If they are post-build, then naturally the parent build is not going to block on them, and thus the result of the parent build is the only thing Jenkins would sent to GitLab. Also, what type of Jenkins job is the parent job?

Simmay93:
@omehegan We trigger them as postbuild steps. Our parent job is a freestyle Jenkins job.

omehegan:
@Simmay93 in that case I would propose you trigger them as build steps, instead of post-build, so you can block the parent job. You can do this using the Parameterized Trigger plugin: https://wiki.jenkins.io/display/JENKINS/Parameterized+Trigger+Plugin

PR #596, opened 2017-07-20
Upgrade to use GitLab APIv4

### LINKED ISSUE #530: Support API v4 (v3 is deprecated)
GitLab has deprecated the API version which is [used by this plugin](https://github.com/jenkinsci/gitlab-plugin/blob/3ea5fad7c5df4e88940ba9e67828a938db1a98a2/src/main/java/com/dabsquared/gitlabjenkins/gitlab/api/GitLabApi.java#L28):

> We will continue to support v3 of the API until August 2017 and so we encourage you to make any necessary changes to applications that use the v3 API.

https://about.gitlab.com/2017/03/22/gitlab-9-0-released/

It should be upgraded to support v4.

ZJvandeWeg:
This commit is the simplest change which seems to suffice to move the
Jenkins plugin to APIv4. [The docs][api-docs] regarding the move from
v3 to v4 and going through most of the source, suggest that is the only
change too.

To summerize, the path changed from `api/v3` to `api/v4` and when
referencing merge requests in paths the IID is now needed instead of the
ID.

Tests are passing, but in this case I feel like that doesn't suggest
anything and this branch will undergo some manual testing soon.

Fixes #530 

[api-docs]: https://docs.gitlab.com/ce/api/v3_to_v4.html

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/gitlab/api/GitLabApi.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/gitlab/api/GitLabApi.java
@@ -25,7 +25,7 @@ import java.util.List;
 /**
  * @author Robin Müller
  */
-@Path("/api/v3")
+@Path("/api/v4")
 public interface GitLabApi {
 
     @POST
@@ -104,18 +104,18 @@ public interface GitLabApi {
     @PUT
     @Produces(MediaType.APPLICATION_JSON)
     @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
-    @Path("/projects/{projectId}/merge_requests/{mergeRequestId}/merge")
+    @Path("/projects/{projectId}/merge_requests/{mergeRequestIid}/merge")
     void acceptMergeRequest(@PathParam("projectId") Integer projectId,
-                            @PathParam("mergeRequestId") Integer mergeRequestId,
+                            @PathParam("mergeRequestIid") Integer mergeRequestIid,
                             @FormParam("merge_commit_message") String mergeCommitMessage,
                             @FormParam("should_remove_source_branch") boolean shouldRemoveSourceBranch);
 
     @POST
     @Produces(MediaType.APPLICATION_JSON)
     @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
-    @Path("/projects/{projectId}/merge_requests/{mergeRequestId}/notes")
+    @Path("/projects/{projectId}/merge_requests/{mergeRequestIid}/notes")
     void createMergeRequestNote(@PathParam("projectId") Integer projectId,
-                                @PathParam("mergeRequestId") Integer mergeRequestId,
+                                @PathParam("mergeRequestIid") Integer mergeRequestIid,
                                 @FormParam("body") String body);
 
     @GET
--- a/src/test/java/com/dabsquared/gitlabjenkins/connection/GitLabConnectionConfigTest.java
+++ b/src/test/java/com/dabsquared/gitlabjenkins/connection/GitLabConnectionConfigTest.java
@@ -71,7 +71,7 @@ public class GitLabConnectionConfigTest {
 
     @Test
     public void doCheckConnection_success() {
-        HttpRequest request = request().withPath("/gitlab/api/v3/.*").withHeader("PRIVATE-TOKEN", API_TOKEN);
+        HttpRequest request = request().withPath("/gitlab/api/v4/.*").withHeader("PRIVATE-TOKEN", API_TOKEN);
         mockServerClient.when(request).respond(response().withStatusCode(Response.Status.OK.getStatusCode()));
 
         GitLabConnectionConfig connectionConfig = jenkins.get(GitLabConnectionConfig.class);
@@ -83,7 +83,7 @@ public class GitLabConnectionConfigTest {
 
     @Test
     public void doCheckConnection_forbidden() throws IOException {
-        HttpRequest request = request().withPath("/gitlab/api/v3/.*").withHeader("PRIVATE-TOKEN", API_TOKEN);
+        HttpRequest request = request().withPath("/gitlab/api/v4/.*").withHeader("PRIVATE-TOKEN", API_TOKEN);
         mockServerClient.when(request).respond(response().withStatusCode(Response.Status.FORBIDDEN.getStatusCode()));
 
         GitLabConnectionConfig connectionConfig = jenkins.get(GitLabConnectionConfig.class);
--- a/src/test/java/com/dabsquared/gitlabjenkins/publisher/GitLabAcceptMergeRequestPublisherTest.java
+++ b/src/test/java/com/dabsquared/gitlabjenkins/publisher/GitLabAcceptMergeRequestPublisherTest.java
@@ -148,7 +148,7 @@ public class GitLabAcceptMergeRequestPublisherTest {
 
     private HttpRequest prepareAcceptMergeRequest(Integer projectId, Integer mergeRequestId) throws UnsupportedEncodingException {
         return request()
-                .withPath("/gitlab/api/v3/projects/" + projectId + "/merge_requests/" + mergeRequestId + "/merge")
+                .withPath("/gitlab/api/v4/projects/" + projectId + "/merge_requests/" + mergeRequestId + "/merge")
                 .withMethod("PUT")
                 .withHeader("PRIVATE-TOKEN", "secret")
                 .withBody("merge_commit_message=Merge+Request+accepted+by+jenkins+build+success&should_remove_source_branch=false");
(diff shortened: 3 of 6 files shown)


ZJvandeWeg:
The first round of manual testing I just concluded, using webhooks as proposed in the readme. Worked well, tomorrow Ill go over the integration GitLab provides.

If anyone wants to help with the manual testing, please feel free! 😄

omehegan:
Thanks very much! I'll publish a snapshot release of the plugin with this patch so others can test it out easily.

omehegan:
The snapshot release with this patch is available for download at http://repo.jenkins-ci.org/snapshots/org/jenkins-ci/plugins/gitlab-plugin/1.4.8-SNAPSHOT/gitlab-plugin-1.4.8-20170720.184525-1.hpi  It can be manually uploaded and installed from the Advanced tab of the Plugin Manager in Jenkins. We would appreciate any testing folks can do, with feedback here.

omehegan:
@ZJvandeWeg is this going to break support for GitLab versions older that 8.17? It's not clear to me when the v4 API was first introduced.

ZJvandeWeg:
@omehegan Older than 9.0. In the major point release APIv4 was introduced, APIv3 is available but deprecated and will be dropped soon.

omehegan:
OK. So we need to make a judgment call about whether we want this to be a breaking plugin release. If this is the whole extent of changes needed for v4, it seems like we could make it a boolean option for folks to enable, so we can still support v3.

ZJvandeWeg:
In my opinion this would bump the version to 1.5.0, so a minor version bump for this plugin. I would expect users to expect changes around a 9.0 release first of all and it will give end users a clear distinction: Use 1.4.X for everything < 9.0. Use 1.5 for anything after that.

Also, this means not maintaining a checkbox to switch the API. To me it seems just not worth it to give the user another choice while actually there is hardly a choice, you either use the new API if its available, or the old one, which don't actually make a difference to user, so why should this be a choice? This also would also allow users to use v3 now, upgrade GitLab to say 9.6 and jenkins stops working for them.

Let me know your thoughts!

ayufan:
How about bumping gitlab-plugin to 2.0.0 to indicate breaking change and advise all people to continue using 1.x if they use 8.x, and use 2.x if they use 9.x?

omehegan:
Yes, a version bump might be a good option. You can also specify that a release is backwards-incompatible, so people see a warning about that in the Jenkins Update Center, and hopefully follow the link from there to see what the change is. That's the best you can do to warn people though. We still always get a few people who blindly upgrade and then wonder why their integration broke :)

ayufan:
We don't really know how many people use with GitLab < 9.0?

omehegan:
@ayufan technically I do ;)  But my old team at work is all being laid off and so their use of GitLab and Jenkins is dormant now. I just help maintain the plugin because I'm a nice guy. If you look through the submitted issues you could get a rough idea of how many people are still on it. In my experience most people are afraid to upgrade a critical system that is working. For better or for worse. I deal with this a lot in the Jenkins world too.

ZJvandeWeg:
@ayufan and @omehegan 

If people are afraid to update, we should just bump it to 2.0 IMO. 😄 

Also, I've done yet another round of testing, didn't spot anything out of order. Did anyone else try the package?

ayufan:
I wonder if we should not consider doing the same for:
https://github.com/Argelbargel/gitlab-branch-source-plugin :)

omehegan:
@ayufan actually @Argelbargel is going to merge the features of that plugin (which was not released publicly AFAIK) into this one.

aundrus:
@omehegan: I tested GitLab plugin version 1.4.8-SNAPSHOT (private-b4157004-omehegan) on our ATLAS CI testbed and it works OK with GitLab version 9.3.9-ee. It works perfectly OK (FYI: ATLAS is the experiment at CERN, ATLAS software workflow is based on the GitLab/Jenkins, with busy CI system running 50 - 100 daily rebuilds on merge requests).

ZJvandeWeg:
Now another (heavy) user verified this patch, I think this PR is ready. I've removed `WIP:` in the title to reflect that state 😉 

Now about the versioning: GitLab will bump the version to 10.0 when APIv3 is actually dropped. This is probably the 22nd of september. My advice would be to bump the version to 2.0 the 22nd too, and let people use 2.0+ with 10.0+. This should be the clearest for everybody.

@omehegan Please let me know if you need anything else before merging this! 😄 

Edit: brief discussion with @ayufan and he suggested that the best route would be to bump to 2.0 right away as it works already and this might prevent broken CI systems.

Argelbargel:
Just an idea, but did you think about supporting both apis for some time? I planned to do that for the gitlab-branch-source-plugin and have some initial (untested) code in place to decide which api to used based on the server-version (but still do not have much time left to really work on it).

Basically all there is to do is to call the api-endpoint for the version once per build, decide which version to use and store this info in a runaction so we do not have to query for it in subsequent api-calls.

Switching between APIs might be a better way than supporting two versions/branches of this plugin in parallel...

adjika:
I have briefly tested it (publish build status integration mostly) and have no issues to report either.

That said, I'd suggest adjusting the docker-gitlab versions in the sample docker-compose.yml file and the pom since they're currently set to 8.x.x

omehegan:
@Argelbargel I do think ideally it would be better to support both API versions simultaneously if possible. From what I've seen of issues being filed, many people are still on 8.x of GitLab. It would be *nice* if we could keep the plugin working for those people and not introduce a hard requirement that they upgrade GitLab in order to keep up with plugin improvements. 

I was not intending to maintain two branches of the plugin; I would simply say that people should not upgrade to 2.0 unless they are on GitLab 9.0+. We can mark the change such that, in the Jenkins update center, they will see a warning about this which leads to the plugin doc for more information.

@ZJvandeWeg what do you think of this? The actual diff to change versions seems so minor that I would think adding an option to switch, and maintain both, would not take much work. Rather than querying GitLab for the version, one could add this as an option to the global GitLab connection configuration. This way, for users who have more than one GitLab instance (rare but supported case), they could specify the version to be used once, and then change it once when they upgrade.

omehegan:
@adjika yes, I have never used the Docker stuff that we include with the plugin source, but I'm going to work on setting it up so I can more easily perform manual testing before releases. I'll work on bringing it up to date.

ZJvandeWeg:
@omehegan I'm not at all comfortable in Java, and not with Jenkins either. So I hesitate a bit to commit to this, also as my time this month is rather limited with the 10.0 release coming in about a month.

If I find the time, I'll take a look, but I doubt that is any time soon.

bikebilly:
@Argelbargel it will be great if you can help on this in order to have both API supported at the same time (via autodetection or an option in the configuration), do you think you can help @ZJvandeWeg with the Jenkins/Java part of the integration? API v3 might be removed really soon and we'd want to guarantee Jenkins users are able to keep everything working while upgrading to the latest version.
Since it is really urgent, I think that the simplest option is the best one, we can also improve it later if needed.

bikebilly:
@omehegan do you think one of the active maintainers of the GitLab plugin can have some time to help here too? It could save us a lot of time!

Argelbargel:
Hi @bikebilly, sorry, i've been away the last days. I'll try to look into it this evening...

Argelbargel:
Hi, i did some preliminary work to support multiple API-implementations over at https://github.com/Argelbargel/gitlab-plugin 

For now, there's just some code in place to support different API-implementations. Next step will be adding the possibility to configure the api-level manually in the UI of GitLabConnection(Config) via a simple select-box. After that i'll see, whether it's possible to auto-detect the api-level supported by the server (via a simple mapping of server-version to api-level)

For now, i made the changes to my own fork and simply copied the changed GitLabApi @ZJvandeWeg  over. Would it be better to fork his version and merge it there first?

Argelbargel:
Hi, i have just finished the work to enable support for both api-versions at https://github.com/Argelbargel/gitlab-plugin.

I am not sure how to proceed, though. Should I open a new merge-request or integrate my changes into @ZJvandeWeg 's branch?

ZJvandeWeg:
I'd be happy to accept such a change!

-- 
Zeger-Jan van de Weg

On 25 August 2017 at 17:28:22, Karsten Kraus (notifications@github.com) wrote:

Hi, i have just finished the work to enable support for both api-versions at https://github.com/Argelbargel/gitlab-plugin.

I am not sure how to proceed, though. Should I open a new merge-request or integrate my changes into @ZJvandeWeg 's branch?

Argelbargel:
Hi @ZJvandeWeg , i'm sorry, but as i forked from jenkins/gitlab-plugin there seems to be no easy way to create a pull-request to this PR. Thus i've create a new one (#614) with me code to switch between your GitLabApiV4-implementation and the current one for GitLabApiV3.

ZJvandeWeg:
@Argelbargel Thats more than ok! I'll close this MR, and hope everyone will continue to discuss at #614 

Thanks for contributing!

Issue #608, opened 2017-08-22
Gitlab Hook names changed


Balint134:
## Issue
Previously working Gitlab configuration stopped working at all, hooks are sent by gitlab but they're not triggering the actual job. After investigation I've found out that gitlab changed Hook headers (without documenting it) thus breaking the plugin's ability to trigger jobs using hooks.

### Context
- **Gitlab plugin version**: 1.4.7
- **Gitlab version**:  9.5.0.RC8
- **Jenkins version**: 2.60.2
- **Job type**: Pipeline

trunov-ms:
"API V3 will be unsupported from GitLab 9.5, to be released on August 22, 2017. It will be removed in GitLab 9.5 or later." 
https://docs.gitlab.com/ce/api/v3_to_v4.html

Balint134:
My main problem is that they're not mentioning this header change anywhere - at least I'm not able to find it.

Feel free to discard the issue and PR if it makes no sense.

kuops:
in gitlab-ce:9.5.1-ce.0  
/project/xxx 
404

eljenso:
@SwankyPantie Could you please link the changes you found? Would be helpful if some other plugin devs stumble upon this problem.

FYI, according to their [press release for the GitLab 9.5](https://about.gitlab.com/2017/08/22/gitlab-9-5-released/#gitlab-ci-api-v1,-gitlab-runner-1.11.x), API v3 will deprecate with GitLab 10.0.

omehegan:
The API version is unrelated to the data that webhooks send, as far as I can guess. We only use the GitLab API to send data *TO* GitLab about build results and so forth. If there has been a breaking change to the hook data that GitLab sends... that's annoying.

@dblessing can you comment on this?

Balint134:
@eljenso I'd be glad to do so, but as I wrote earlier there seems to be no documentations related to these changes. All I can see is that the "Test webhook" fails because of the Header changes.

omehegan:
@SwankyPantie I assume you upgraded GitLab and then this broke. What version did you upgrade from? We can probably find the commit where this was changed, and maybe some more information on it.

Balint134:
@omehegan I'm using the hosted `gitlab.com` version. According to source code this is the snippet doing the magic: [link](https://gitlab.com/gitlab-org/gitlab-ce/blob/master/app/services/web_hook_service.rb#L114) but this file hasn't been changed for a while.
[...]

PR #609, opened 2017-08-22
Add missing gitlab hook event headers

Balint134:
PR for https://github.com/jenkinsci/gitlab-plugin/issues/608

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/webhook/ActionResolver.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/webhook/ActionResolver.java
@@ -106,13 +106,18 @@ public class ActionResolver {
         String tokenHeader = request.getHeader("X-Gitlab-Token");
         switch (eventHeader) {
             case "Merge Request Hook":
+            case "Merge Request Event":
                 return new MergeRequestBuildAction(project, getRequestBody(request), tokenHeader);
             case "Push Hook":
+            case "Push Event":
             case "Tag Push Hook":
+            case "Tag Push Event":
                 return new PushBuildAction(project, getRequestBody(request), tokenHeader);
             case "Note Hook":
+            case "Note Event":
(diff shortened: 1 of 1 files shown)


omehegan:
Thanks for the bug report and this PR. But if I merge this as-is, presumably it's going to break anyone using an earlier version of GitLab. I need some feedback from the GitLab folks about what happened here. I've requested that on the Issue page.

adjika:
@omehegan I can't speak for the PR since I don't have a GitLab 9.5.0 at hand to verify the issue right now, but could you clarify why the PR breaks older versions? From my, admittedly limited, understanding of the plugin it should still work with older GitLab versions since he just added fallthrough cases for the new x-gitlab-events, so I feel like I'm missing something here.

Balint134:
@omehegan It should be backward compatible, as it is only extending the cases with fall-through options. Yeah feedback would be awesome from Gitlab because I can not find any documentation related to these changes, all I can see that the "Test webhook" fails thus of this Header change.

Argelbargel:
@omehegan: as the header X-Gitlab-Event is only used by gitlab-plugin in the method ActionResolver#onPost which @SwankyPantie has changed and the header is not passed along or processed any further, this should indeed not break compatiblity.

omehegan:
Oh, I'm sorry, I misread the diff; you're not removing the old entries, just adding these additional ones. I get it now.

omehegan:
@SwankyPantie can you resolve merge conflicts?

Balint134:
@omehegan Sure, done

omehegan:
This appears to be a bug that was introduced in GitLab. For more context, please see the comments in the #608 issue. I am leaning towards not merging this, and letting GitLab fix it on their side, because this only affects the Test Hook functionality.

omehegan:
Bug was fixed on GitLab side, see #608.

PR #611, opened 2017-08-28
Adds support for travis-builds and maven-wrapper

Argelbargel:
This adds basic support for travis-ci which enables all forks to quickly and easly run all tests (including the integration tests) without having to install docker on their machine and/or creating a pull request.

The included travis-file simply runs the unit-tests & findbugs and in a parallel job it runs the integration-tests. For an example see: https://travis-ci.org/Argelbargel/gitlab-plugin/builds/269170885

The travis-build also caches maven-artifacts and docker-images to speed up building.


omehegan:
Thanks for doing this work! Although I wonder if it would be more appropriate to build the plugin in Jenkins :)  Let me just see if there is a way we could do that and achieve a similar result.

Argelbargel:
Of course it would be more appropriate. travis-ci just has the advantage of it's ootb-integration with github. If there's a cloudbees or some other provider provides a similar Jenkins-based ci-service which we might integrate, I'll be happy to write a Jenkinsfile for that... otoh, as we basically just have to run maven-targets it doesn't really matter what we use, as long as it makes it really easy to run the tests so they're run more often and break less. It's a bit like hosting the gitlab-plugin on github ;-)

omehegan:
@Argelbargel https://github.com/jenkinsci/gitlab-plugin/commit/c6bd0e2eefd80e5d414c3a82be586ee529307e1e :)

Argelbargel:
As far as I can see, ci.jenkins.io only runs builds for branches and PRs opened on the main project, not for changes made to forks? How would I (or any other forker) run the tests before opening a PR, including the integration-tests (besides locally which blocks at least my machine for about 15 minutes)?

omehegan:
@Argelbargel that is true. I think the best practice in that case would be to simply open a PR marked WIP, for 'Work in Progress.' Then the tests will be run on ci.jenkins.io each time you push to the PR branch. Once you're satisfied with the PR, you can remove the WIP and we will know it's ready for review and merge.

If you want to see more of what happens during the build on CI, you can see the code here: https://github.com/jenkins-infra/pipeline-library

Issue #613, opened 2017-08-28
Jenkins jobs defined using generated DSL must specify all possible configuration for triggers, publishers, etc


g-dx:
## Issue
Jenkins jobs defined using the `gitlab-plugin` generated DSL must specify all configuration. 

For example:

```
job('job-name') {

    scm {
        git('...some repo...')
    }

    triggers {
        gitlab {
            triggerOnPush(true)
            branchFilterType('NameBasedFilter')
            includeBranchesSpec 'master'
            triggerOnMergeRequest(false)
            triggerOnAcceptedMergeRequest(false)
            triggerOnClosedMergeRequest(false)
            triggerOpenMergeRequestOnPush('never')
            triggerOnNoteRequest(false)
            noteRegex('#rebuild')
            setBuildDescription(true)
            addNoteOnMergeRequest(false)
            skipWorkInProgressMergeRequest(true)
            ciSkip(true)
            addCiMessage(false)
            addVoteOnMergeRequest(false)
            acceptMergeRequestOnSuccess(false)
            excludeBranchesSpec('')
            targetBranchRegex('')
            mergeRequestLabelFilterConfig {
                include('')
                exclude('')
            }
            secretToken('')
        }
    }
    
    publishers {
        gitLabCommitStatusPublisher {
            name 'jenkins'
            markUnstableAsSuccess false
        }
        gitLabMessagePublisher {
            onlyForFailure true
            replaceSuccessNote false
            replaceFailureNote false
            replaceAbortNote false
            successNoteText ''
            failureNoteText ''
            abortNoteText ''
        }
    }

    steps {
        /* ... some steps ... */
    }
}

```
Otherwise executing the DSL script will fail.

For example the code above will fail when run against v1.4.7 of the `gitlab-plugin` with the following error:

```
Build step 'Invoke Gradle script' changed build result to SUCCESS
Processing DSL script job.groovy
ERROR: (job.groovy, line 8) the following options are required and must be specified: triggerOnAcceptedMergeRequest, triggerOnClosedMergeRequest
```
Due to the changes introduced in https://github.com/jenkinsci/gitlab-plugin/commit/a61a6c77be85b89a5f8b6d37cc382fb474ea5262.

Currently this means all new additions to the plugin DSL will cause _existing_ jobs to break when the `gitlab-plugin` is upgraded.

Is this a limitation of Jenkins when leveraging generated DSL functionality or can a feature request be considered to support using default values for configuration which is not specified?

### Context
[...]

PR #614, opened 2017-08-28
Adds support for GitLabAPI V4 while keeping support for V3

Argelbargel:
This is an enhancement of PR #596 which adds support for gitlab-api v4 while keeping support for v3
(as i forked jenkins/gitlab-plugin i could not find a simple way to add to @ZJvandeWeg 's PR). 
GitLabConnectionConfig has a new configuration option "API-Level" in the advanced configuration section where you can switch between v3 and v4. Default is "autodetect" which tries to determine which api-endpoint is available (v3 is preferred for now). The result of the autodetection is cached after the first request to the server.

Additionally, i've tried to change the name of the classes so that the class-names correspond with the names of the getters and variables, thus GitLabApi is now called GitLabClient.

This is just some fancy code to switch the interface-definitions used to create a gitlab-rest-client. i'm not sure if switching between apis will work for existing jobs which do not have both the id and iid of merge-requests stored somewhere - could be that we have to change other parts to build a seamless upgrade path to v4.

PR #616, opened 2017-08-29
Support @DataBoundSetter for triggers and publishers

g-dx:
Currently jenkins jobs defined using the generated DSL for the `gitlab-plugin` are required to supply all possible values for any defined triggers or publishers due to the use of the `@DataBoundConstructor` annotation. This leads to the problem described in #613 where any modifications to the trigger or publisher constructor signatures cause existing job definitions to fail when run against a new version of the plugin.

Fix this by adding support for a setter driven approach to configure triggers and publishers with the `@DataBoundSetter` annotation. This allows all unimportant configuration to be ignored when defining jobs and prevents new functionality from breaking existing jobs when upgrading.

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/GitLabPushTrigger.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/GitLabPushTrigger.java
@@ -48,6 +48,7 @@ import org.jenkinsci.Symbol;
 import org.kohsuke.stapler.Ancestor;
 import org.kohsuke.stapler.AncestorInPath;
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
 import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.Stapler;
 import org.kohsuke.stapler.StaplerRequest;
@@ -76,9 +77,9 @@ public class GitLabPushTrigger extends Trigger<Job<?, ?>> {
     private boolean triggerOnMergeRequest = true;
     private boolean triggerOnAcceptedMergeRequest = false;
     private boolean triggerOnClosedMergeRequest = false;
-    private final TriggerOpenMergeRequest triggerOpenMergeRequestOnPush;
+    private TriggerOpenMergeRequest triggerOpenMergeRequestOnPush;
     private boolean triggerOnNoteRequest = true;
-    private final String noteRegex;
+    private String noteRegex = "";
     private boolean ciSkip = true;
     private boolean skipWorkInProgressMergeRequest;
     private boolean setBuildDescription = true;
@@ -91,7 +92,7 @@ public class GitLabPushTrigger extends Trigger<Job<?, ?>> {
     private String includeBranchesSpec;
     private String excludeBranchesSpec;
     private String targetBranchRegex;
-    private final MergeRequestLabelFilterConfig mergeRequestLabelFilterConfig;
+    private MergeRequestLabelFilterConfig mergeRequestLabelFilterConfig;
     private volatile Secret secretToken;
 
     private transient BranchFilter branchFilter;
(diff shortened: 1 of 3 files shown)


src/main/java/com/dabsquared/gitlabjenkins/GitLabPushTrigger.java
```diff
@@ -103,9 +103,10 @@
 
 
     @DataBoundConstructor
-    @GeneratePojoBuilder(intoPackage = "*.builder.generated", withFactoryMethod = "*")
```
christ66: I'm not familiar with this annotation, however I believe this is generating code which might break backward compatibility.

src/main/java/com/dabsquared/gitlabjenkins/GitLabPushTrigger.java
```diff
@@ -103,9 +103,10 @@
...
-    public GitLabPushTrigger(boolean triggerOnPush, boolean triggerOnMergeRequest, boolean triggerOnAcceptedMergeRequest, boolean triggerOnClosedMergeRequest, 
-    						 TriggerOpenMergeRequest triggerOpenMergeRequestOnPush, boolean triggerOnNoteRequest, String noteRegex, 
+    public GitLabPushTrigger() { }
+
+    public GitLabPushTrigger(boolean triggerOnPush, boolean triggerOnMergeRequest, boolean triggerOnAcceptedMergeRequest, boolean triggerOnClosedMergeRequest,
```
christ66: I would recommend deprecating the constructors and adding a javadoc to recommend using getter/setter methods.

src/main/java/com/dabsquared/gitlabjenkins/publisher/GitLabMessagePublisher.java
```diff
@@ -35,7 +36,6 @@
     private String abortNoteText;
     private String unstableNoteText;
 
-    @DataBoundConstructor
     public GitLabMessagePublisher(boolean onlyForFailure, boolean replaceSuccessNote, boolean replaceFailureNote, boolean replaceAbortNote, boolean replaceUnstableNote,
```
christ66: Probably best to deprecate the constructor.

src/main/java/com/dabsquared/gitlabjenkins/GitLabPushTrigger.java
g-dx: Good point. Will leave as is.

src/main/java/com/dabsquared/gitlabjenkins/GitLabPushTrigger.java
g-dx: Done.

src/main/java/com/dabsquared/gitlabjenkins/publisher/GitLabMessagePublisher.java
g-dx: Done.

g-dx:
PTAL.

omehegan:
@christ66 can you take another look at this one?

src/main/java/com/dabsquared/gitlabjenkins/GitLabPushTrigger.java
```diff
@@ -101,11 +102,11 @@
     private transient boolean acceptMergeRequestOnSuccess;
     private transient MergeRequestLabelFilter mergeRequestLabelFilter;
 
-
+    @Deprecated
```
christ66: Did something happen to the empty constructor you created? I was only mentioning to use the `@deprecated` annotation only if there was a new constructor which should be used instead of this one, and this one should be deprecated in favor of the new constructor + using setter methods. If the other empty constructor is not needed, then you should remove the deprecated annotation. I also believe that adding a `@deprecated` should require javadoc suggesting to use the other constructor.

g-dx: Ah, apologies. I stripped the merge back to only what's required to support setters and left the rest for fear of affecting backwards compatibility.

I will add no-arg constructors where necessary and update `@Deprecated` constructors to inform clients to use the no-arg version in conjunction with setters.

src/main/java/com/dabsquared/gitlabjenkins/publisher/GitLabMessagePublisher.java
```diff
@@ -35,6 +36,10 @@
     private String abortNoteText;
     private String unstableNoteText;
 
+    /**
+     * @deprecated use {@link #GitLabMessagePublisher()} with setters to configure an instance of this class.
+     */
+    @Deprecated
     @DataBoundConstructor
```
christ66: Should the annotation `@DataBoundConstructor` be around the `public GitLabMessagePublisher() { }` constructor instead of the current constructor.

g-dx: I must admit I'm somewhat unsure exactly where `@DataBoundConstructor` is actually required now that `@DataBoundSetter` has been added. 

When making these changes I used [this](https://github.com/jenkinsci/slack-plugin/pull/236/files) merge from the `slack-plugin` as a reference. Looking again at those changes they do appear to have moved `@DataBoundConstructor` annotation onto the no-arg constructors. As such I will make the same change.

christ66 (approved):
LGTM 👍

g-dx:
@christ66, @omehegan - if these changes are ok any ETA on when these changes will find their way into a release?

omehegan:
@g-dx right now I am holding releases until we can get the GitLab v4 API support complete. That is being tracked in #614.

g-dx:
No problem. Thanks for the update.

omehegan:

jakub-bochenski:
Doesn't seem to work, created #674

Issue #499, opened 2017-02-10
Consider implementing branch source functionality


omehegan:
When using Jenkins with GitHub, it is now possible to create something called a 'GitHub org folder,' which is a 'smart folder' that monitors your entire GitHub org. Anytime a new repo is created in GitHub, with a branch that has a Jenkinsfile in it, Jenkins automatically creates a new Pipeline multibranch job and sets up appropriate triggers. This used to be called GitHub Org Folder, it's now called GitHub Branch Source.

@kohsuke has written a new plugin which depends on gitlab-plugin and provides this functionality: https://github.com/kohsuke/gitlab-branch-source-plugin  But maybe we should consider supporting it natively in this plugin. It's not clear to me how much more work would be required to do that.

@trance1st @colourmeamused do you want to take a look at this and see what you think? It's beyond my ability, but I know there is a lot of excitement about GitHub Org Folder, and it would be good if we can try to keep parity with things that are supported with Jenkins + GitHub.

kohsuke:
The plugin Owen linked implements everything that's necessary to make "GitLab org folder" work already. You can try it with `mvn hpi:run`.

What I'd like you guys to consider is whether we can "merge" two plugins by taking code in that repo and putting it here. From users' perspective I think it makes the most sense, because only a handful people in the world would be able to see the name "gitlab branch source plugin" and understand what it adds to your Jenkins.

Owen told me that this plugin used to use the java-gitlab-api that I'm using in gitlab-branch-source-plugin but since then moved away from it. I explained to him that my priority is to get this functionality delivered to users as fast as possible, and I don't have any attachment to the code. So if you decide to eliminate that dependency over time I'm perfectly happy, but I'm not very keen to doing that myself. I can imagine things like this might cause some hesitation to merge plugins, in which case I'm happy to keep the gitlab-branch-source plugin separate.

What I think is a great tragedy is for this plugin to "natively support it" by essentially reimplementing the same feature in a different way, which results in duplicated functionality that spreads us even thinner.

omehegan:
@coder-hugo if you have a couple of minutes, can you refresh my memory about why we decided to stop using the java-gitlab-api library, and instead just implement our own access to GitLab? IIRC there was a bug or missing functionality in that library, and you concluded that it was pretty easy to just call it directly ourselves.

Argelbargel:
Hi there,

[...]

PR #621, opened 2017-09-05
WIP: Push to target branch should rebuild merge requests

### LINKED ISSUE #620: Push to target branch should rebuild merge requests
## Feature request
I have a multi branch pipeline project and want to build all branches if someone pushes to the target branch.

The story behind it:
In branch driven development for each new feature, bugfix and so on a new branch is created. For a finished branch a merge request is created and it can't be merged until the Jenkins has completed the build. So there are two conditions before merging, manual review and successful Jenkins build.

After merging one branch it should be enforced to rebuild all other branches having the same target branch, f.e. feature-1 and feature-2 should be merged into develop. feature-1 is merged and feature-2 should be automatically rebuild to ensure that it is compatible to the new changes and would not break after merge the build of develop. During the rebuild, the develop is merged into the feature,

CSchulz:
This PR is work in progress and base for discussions.

I have added some TODOs to the code to hightlight potential stuff for discussion.

Fixes #620

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -192,6 +192,11 @@
         <artifactId>matrix-project</artifactId>
         <version>1.10</version>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>branch-api</artifactId>
+      <version>2.0.11</version>
+    </dependency>
 
     <!-- REST client dependencies -->
     <dependency>
--- a/src/main/java/com/dabsquared/gitlabjenkins/webhook/build/PushBuildAction.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/webhook/build/PushBuildAction.java
@@ -1,13 +1,18 @@
 package com.dabsquared.gitlabjenkins.webhook.build;
 
 import com.dabsquared.gitlabjenkins.GitLabPushTrigger;
+import com.dabsquared.gitlabjenkins.connection.GitLabConnectionProperty;
+import com.dabsquared.gitlabjenkins.gitlab.api.GitLabApi;
+import com.dabsquared.gitlabjenkins.gitlab.api.model.MergeRequest;
 import com.dabsquared.gitlabjenkins.gitlab.hook.model.Project;
 import com.dabsquared.gitlabjenkins.gitlab.hook.model.PushHook;
+import com.dabsquared.gitlabjenkins.gitlab.hook.model.State;
 import com.dabsquared.gitlabjenkins.util.JsonUtil;
 import hudson.model.Item;
 import hudson.model.Job;
 import hudson.security.ACL;
 import hudson.util.HttpResponses;
+import jenkins.branch.MultiBranchProject;
 import jenkins.model.Jenkins;
 import jenkins.plugins.git.GitSCMSource;
 import jenkins.scm.api.SCMSource;
@@ -15,9 +20,13 @@ import jenkins.scm.api.SCMSourceOwner;
 import org.apache.commons.lang.StringUtils;
 import org.eclipse.jgit.transport.URIish;
 
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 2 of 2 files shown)


Argelbargel:
Hi @CSchulz, could you describe why you would like to build all other branches when there is a push to the master (or the he target-branch)? In most setup those other branches should not have any changes in them or am I missing something?

[EDIT] Ah, if I'm not mistaken you're code rebuilds all merge-requests when there are changes to the branch they should be merged into, right?

CSchulz:
I have added the story in the issue description.

Argelbargel:
Ok, I've read the issue this afternoon, when the use case was not yet there, sorry. 

I'd prefer a configuration option ("build on pushes to the target branch") for the trigger though, to make it explicit that pushes to the target-branch trigger builds in the mr-jobs.

CSchulz:
Yes of course, the PoC was just for showing it can work, but I wasn't sure how to implement the configuration stuff.

Reusing the existing GitlabPushTrigger would not make so much sense, because only one option would be added and is interesting for the configuration, if I am right.

What about GitlabMultiBranchTrigger?

CSchulz:
@coder-hugo @omehegan 

I am thinking about the configuration of the trigger for a job. As far as I understood the code and documentation correctly the usage of gitlab as trigger does not work with a multibranch project, right?

Shall I introduce a new configuration trigger or a new property?
Or better use the existing GitlabPushTrigger and don't support every option?

~I think it is a good idea to extend the existing AbstractWebHookTriggerHandler to get the commit status update and so on. Furthermore to handle additional configuration options like mergeRequestLabelFilte. ciSkip and branchFilter seems not to be needed in this use case.~
This would make sense if we are reusing the GitlabPushTrigger, because it is registering the GitLabCommitStatusPublisher, which is used by the abstract handler.

omehegan:
@CSchulz our understanding of the best practice for implementing multibranch support was that it shouldn't need "triggers" to be configured; GitLab just triggers the job, and Jenkins takes care of indexing the branches and building the ones that have changed. So that is how this support was written. I would need someone with more experience implementing multibranch to comment on this I guess.

(FWIW, @coder-hugo is no longer an active maintainer)

pom.xml
```diff
@@ -192,6 +192,11 @@
         <artifactId>matrix-project</artifactId>
         <version>1.10</version>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>branch-api</artifactId>
```
jglick, line 197: No!

src/main/java/com/dabsquared/gitlabjenkins/webhook/build/PushBuildAction.java
```diff
@@ -102,7 +111,64 @@ public void run() {
                     }
                 }
             }
+
+            // TODO configuration can be done via "trigger" like the existing GitlabPushTrigger or similar object
+            // TODO should be replaced with getClass comparison to prevent the dependency to the branch api
+            if (project instanceof MultiBranchProject) {
```
jglick, line 117: This not the right approach. You “just” need to make sure your `SCMRevision` implementation encodes the right things. Look at `github-branch-source` for an example.

jglick (changes requested):
And see discussion in JENKINS-37491.

Argelbargel:
Hi, The problem is, that the project-trigger-based-approach of the gitlab-plugin currently does not really match the event-based approach of scanning for projects to build that the multibranch-projects use.

Have a look at at the code of https://github.com/jenkinsci/gitea-plugin for a reference-implementation for integrating something like gitlab with multibranch-projects.

@omehegan: the whole day i've been thinking about how to comment that this feature would better, cleaner and more in line with the concepts of the branch- and scm-apis be implemented into the [gitlab-branch-source-plugin](https://github.com/Argelbargel/gitlab-branch-source-plugin) - without getting into when i will find the time to merge that plugin with this one (which i still really, really like to do but life and payed work keeps getting in the way) ;-)

So @CSchulz, if you'd be willing to use a plugin that's not "offically" released and which will be deprecated and merged with this one in the next months, i'll try to get your use-case working in the gitlab-branch-source-plugin (or you could try for yourself). As i said, this should be easy to do compared to a proper integration of your pr with the current gitlab-plugin - in the end you'd have to rewrite much of the gitlab-branch-source-plugin's code for proper multibranch-support. 

Oh, and if you'd be interested to help integrating the gitlab-branch-source-plugin with this one, that would be very welcome ;-)

CSchulz:
Thanks for all the hints, I will try to make the best out of it.

Did I understand you correctly, the plan is to integrate the gitlab-branch-source-plugin into this one?
And it would be better to integrate my idea into it before?

Argelbargel:
In my opinion this would be the quickest and easiest way to make your use-case work. But i am a bit biased of course ;-)

See #499 for the discussion about integrating both plugins

src/main/java/com/dabsquared/gitlabjenkins/webhook/build/PushBuildAction.java
CSchulz: Thanks for the hints. It is great to see people developing Jenkins here also.

The idea of my use case is to rebuild branches in a multibranch project if there merge target changes.
So I have to track the target branch rev within the SCMRevision implementation to be able to see if a new build is necessary when I trigger the branch indexing?

omehegan:
@CSchulz @Argelbargel yes, I chatted with @jglick about this PR on IRC, which led to his comments here. His comment to me was that this feature is essentially implemented in github-branch-source, so we might look to that for an example. And then that reminded me that @Argelbargel has a WIP to implement branch-source functionality here in the gitlab-plugin.

I'm all for having this feature, I just want to make sure we follow best practices when implementing it. And @jglick is definitely one of the primary arbiters of what those are when it comes to Pipeline :)

omehegan:
Should we close this PR?

CSchulz:
I think it is okay, I have no time for diving into the code of @Argelbargel. After it is merged this feature would be easy to integrate. The issue is still open.

PR #628, opened 2017-09-15
make it clear that the note is a regex, not a plain string

evgeni:
otherwise people might enter '[test]' and wonder why a '[test]' comment does not trigger a build.


omehegan:
Thanks, this is a good catch!

PR #632, opened 2017-09-26
Allow to decide on the change-set whether a build should be triggered in case of a push-/pipeline event

SourcePond:
There are cases where more than one job is necessary to build different parts of the same GIT-project. Suppose a multi-module project with an API- and an implementation module. In such a situation you may not want to build the API module in case when only the implementation module has been changed. Therefore, this pull-request introduces a new feature which makes it possible to specify an inclusion regex on a job (advanced configuration). If specified, the job will only be triggered via Web-Hook, when the change-set (add/modify/remove) contains at least one file which matches the specified regular expression. This means, you can specify for each module a job with a specific regex, and, an appropriate Web-Hook call on GitLab.

To achieve this, following changes had been made on the original code:

- Change the names of existing classes to more generic counterparts
- Introduced field "includeFilesRegex" on config page (advanced section)
- Introduced new factory method for file-filters
- Enhanced WebHookTriggerHandler::handle so it can retrieve the file-filter
- Implemented decision logic whether to start a job on AbstractWebHookTriggerHandler


omehegan:
@SourcePond can you resolve the merge conflicts here? Then please ping @Argelbargel and @mreichel for review. 

Actually, I want to add: I have not had a chance to look at the diff and won't have time myself just now, but can you ensure that a notice will be logged when the plugin does not trigger a build due to this feature (if you haven't already done that)? I just want to make sure we don't make it more difficult to debug since, "Why didn't the plugin trigger my build?" is one of the more common support requests we get. Thanks!

omehegan:
@SourcePond do you still want to try to get this merged?

SourcePond:
@omehegan I'm to busy to have a look at it right now; I may open a new pull request in near future.

PR #641, opened 2017-10-16
Added new hook events sent by Gitlab starting in 9.5.4

jaraujoduarte:
Since the 9.5.4 upgrade of Gitlab, when testing the hooks from the "integrations" section, the "X-Gitlab-Event" with the event description is sent in the form "xxx Event" (i.e. "Push Event") instead of the regular "xxx Hook", showing a 404 error on the Gitlab side when the hook is actually working as Gitlab does send "xxx Hook" kind of headers when the actual event is triggered.

Although it looks like a bug/inconsistency from Gitlab, this change extends the list of supported headers to overcome this inconsistency.


DieBauer:
Duplicate of #609?

omehegan:
Bug was fixed on GitLab side, see #608.

PR #649, opened 2017-10-26
Bugfix & Test for #648

Argelbargel:
Fixes #648. Sorry for the stupid mistake :-/

PR #650, opened 2017-10-26
Improved gitlab api autodetection

Argelbargel:
Based on the comments/suggestions from @bikebilly in #614 i've changed the order in which the apis are detected and improved the autodetection of the api-level so it does not require a restart of jeninks when the gitlab-instance used stops supporting v3.

This comes at a cost though:
- the code is not really easier to read (might improve, once we do not have support older jvm versions)
- if there really is an error (e.g. gitlab-project was deleted), we're re-trying to detect the api-level before acknowleding it's no use


omehegan:
@Argelbargel nice work :)

Since April of this year Jenkins has only supported Java 8. If we consider only supporting Java 7 or newer in the plugin, would that make your code any cleaner?

Argelbargel:
@omehegan: no, Java8's functional interfaces, method references and lambda expressions will make it a bit more concise (or rather: less verbose). But that'll have to wait some months or so.

omehegan:
@Argelbargel hmm, when I try to build this to release from my laptop, this new test fails:
```
    [ERROR] /Users/omehegan/source/gitlab-plugin/src/test/java/com/dabsquared/gitlabjenkins/gitlab/api/impl/ResteasyGitLabClientBuilderTest.java:[31,45] constructor ResteasyGitLabClientBuilder in class com.dabsquared.gitlabjenkins.gitlab.api.impl.ResteasyGitLabClientBuilder cannot be applied to given types;
      required: java.lang.String,int,java.lang.Class<? extends com.dabsquared.gitlabjenkins.gitlab.api.GitLabApi>
      found: java.lang.String,java.lang.Class<com.dabsquared.gitlabjenkins.gitlab.api.impl.V3GitLabApiProxy>
      reason: actual and formal argument lists differ in length
```

Argelbargel:
Oh. I'll look into it. You're using jdk6 to build?

omehegan:
@Argelbargel no: `java version "1.8.0_74"` On MacOS.

Argelbargel:
Ok. I can reproduce this. Funny enough, the unmerged head of this pr builds without the problem... Probably some change caused by the merge. I'll investigate

omehegan:
Cool, if you can get it fixed I will do a release today.

Argelbargel:
The problem is not directly caused by this pr but by #649. The new code from there does not yet know about the new constructor. Easily fixed. should i push to master or create another pr?

omehegan:
Just push to master.

Argelbargel:
Done. But you might want to hold off the release for tomorrow. I'll try to fix #652, look into #655 and review #653 tonight...

omehegan:
OK, I will hold off.

Argelbargel:
Hrmpf. Fixing #652 will take me a bit longer as i'll have to move the code selecting the correct merge-request-id to use into the api-implementation :-/

omehegan:

PR #653, opened 2017-10-31
Fix unsupported date format in merge request trigger

TheUbuntuGuy:
GitLab now submits new formats for dates on the MR trigger. This change prevents the plugin from crashing when receiving such a trigger from a new version of GitLab. Old formats are preserved for compatibility.


Argelbargel:
@omehegan: is ist possible to restart the checks without commiting to the pr's branch? if think, this should build now that i've fixed the compilation error

TheUbuntuGuy:
I just rebased the branch on the latest master which kicked off a test.

omehegan:

PR #656, opened 2017-11-02
Fixes #652

### LINKED ISSUE #652: 1.5.0 fails to accept merge requests when using the API V4
### Context
- **Gitlab plugin version**: 1.5.0
- **Gitlab version**: 9.5
- **Jenkins version**: 2.73.2

### Logs & Traces

```
Oct 30, 2017 2:22:40 PM
com.dabsquared.gitlabjenkins.publisher.GitLabAcceptMergeRequestPublisher
perform
SEVERE: Failed to accept merge request for project '8348'
javax.ws.rs.NotFoundException: HTTP 404 Not Found
```

### Problem description

We have the impression that when asking the plugin to use the V4 API, at the time of merging a merge request, the ID of the MR passed via the URL is the global one, making the request fail.

This looks correct in terms of how the URL is composed and the variables used:

```java

    @PUT
    @Produces(MediaType.APPLICATION_JSON)
    @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
    @Path("/projects/{projectId}/merge_requests/{mergeRequestIid}/merge")
    @Override
    void acceptMergeRequest(@PathParam("projectId") Integer projectId,
                            @PathParam("mergeRequestIid") Integer mergeRequestIid,
                            @FormParam("merge_commit_message") String mergeCommitMessage,
                            @FormParam("should_remove_source_branch") boolean shouldRemoveSourceBranch);
```

However the value of ```mergeRequestIid``` is not the internal identifier but the global one. The GitLab logs kind of confirm this theory.

Could you please try to reproduce this behaviour?

Thanks.

Argelbargel:
This fixes #652. To fix this, i've had to change the public API of GitLabClient. The methods to accept a merge-request or create a note for a merge-request now require a MergeRequest object instead of project- and merge-request-id. The implementations of GitLabClient extract the proper id required by the api-calls. Furthermore GitLabClient is now just an interface which is implemented by the classes in the impl-package.
As this break api-compatiblity for users of the GitLabClient anyway, i've moved the interface GitLabApi so that GitLabClient is the sole interface known to clients outside of the plugin and we may change the internals without further breaking changes. Might be good to have @mreichel  or @christ66 to sign off on this, though ;-) 

Oh, and i changed GitLabRule so that implementation-classes do not have to check whether the jenkins-instance is already initialized (non-null) any longer.

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/cause/CauseData.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/cause/CauseData.java
@@ -1,5 +1,6 @@
 package com.dabsquared.gitlabjenkins.cause;
 
+import com.dabsquared.gitlabjenkins.gitlab.api.model.MergeRequest;
 import hudson.markup.EscapedMarkupFormatter;
 import jenkins.model.Jenkins;
 import net.karneim.pojobuilder.GeneratePojoBuilder;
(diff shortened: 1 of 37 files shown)


mreichel (commented):
It seems to be working fine and I don't know how to fix the bug without changing the api, so I think we have to go with breaking compatibility

omehegan:

PR #659, opened 2017-11-06
Trigger builds when MR is approved

jakub-bochenski:
The builds are not triggered because parsing of the action field fails.

I can't find this field being actually read anywhere,  hence no automated tests.

I verified manually that this works using our staging Jenkins instance and gitlab  10.0.4-ee


marcinmantke:
Is there any ETA when this will be merged and released?

omehegan:
Thanks!

omehegan:

PR #680, opened 2017-12-06
Removed the build number from target url when the state is pending

### LINKED ISSUE #525: Wrong target_url for pending status when parallell builds are queued
### Context
- **Gitlab plugin version**: 1.4.5
- **Gitlab version**: 8.17.2-ee
- **Jenkins version**: 1.650
- **Job type**: Maven project

### Logs & Traces
Creation of pending commit status. The target_url does not point to the build that actually built branch-A later. That was in fact done by build 680.
```
Call GitLab:
HTTP method: POST
URL: https://<gitlab-server>/api/v3/projects/1/statuses/d516fe7776d11d4bda4e9002e3a8c78b19fb90e5
Request headers: [
Accept = [application/json],
Content-Type = [application/x-www-form-urlencoded],
]
Mar 15, 2017 1:07:37 PM FINEST com.dabsquared.gitlabjenkins.gitlab.GitLabClientBuilder
Got response from GitLab:
URL: https://<gitlab-server>/api/v3/projects/1/statuses/d516fe7776d11d4bda4e9002e3a8c78b19fb90e5
Status: 201 Created
Response headers: [
Cache-Control = [max-age=0, private, must-revalidate],
Connection = [keep-alive],
Content-Length = [557],
Content-Type = [application/json],
Date = [Wed, 15 Mar 2017 12:07:22 GMT],
Etag = [W/"be7adc8781d5ad02b00fd4e2a1a46485"],
Server = [nginx],
Vary = [Origin],
X-Request-Id = [fa413498-6444-4090-ae44-52d00122b0b9],
X-Runtime = [0.154868]
]
Response body: {
  "id" : 1203,
  "sha" : "d516fe7776d11d4bda4e9002e3a8c78b19fb90e5",
  "ref" : "branch-A",
  "status" : "pending",
  "name" : "jenkins",
  "target_url" : "https://<jenkins-server>/jenkins/job/GitlabMergeRequests/679/",
  "description" : null,
  "created_at" : "2017-03-15T12:07:22.184Z",
  "started_at" : null,
  "finished_at" : null,
  "allow_failure" : false,
  "author" : {
    "name" : "Administrator",
    "username" : "root",
    "id" : 1,
    "state" : "active",
    "avatar_url" : "https://<gitlab-server>/uploads/user/avatar/1/avatar.png",
    "web_url" : "https://<gitlab-server>/root"
  }
}
```

Creation of running commit status. Same target_url, but different commit and branch. However, this time it is correct. Build 679 did build that branch and commit.
```
Mar 15, 2017 1:14:56 PM FINEST com.dabsquared.gitlabjenkins.gitlab.GitLabClientBuilder
Call GitLab:
HTTP method: POST
URL: https://<gitlab-server>/api/v3/projects/1/statuses/9d0dec909237cb2c650c7a9f90797d6952da37a4
Request headers: [
Accept = [application/json],
Content-Type = [application/x-www-form-urlencoded],
]
Mar 15, 2017 1:14:56 PM FINEST com.dabsquared.gitlabjenkins.gitlab.GitLabClientBuilder
Got response from GitLab:
URL: https://<gitlab-server>/api/v3/projects/1/statuses/9d0dec909237cb2c650c7a9f90797d6952da37a4
Status: 201 Created
Response headers: [
Cache-Control = [max-age=0, private, must-revalidate],
Connection = [keep-alive],
Content-Length = [533],
Content-Type = [application/json],
Date = [Wed, 15 Mar 2017 12:14:46 GMT],
Etag = [W/"5f7017b8d76bc00e6d31e5d620b8bd07"],
Server = [nginx],
Vary = [Origin],
X-Request-Id = [30574f32-9415-456e-ab77-1feaa3c7dbdb],
X-Runtime = [0.140720]
]
Response body: {
  "id" : 1204,
  "sha" : "9d0dec909237cb2c650c7a9f90797d6952da37a4",
  "ref" : "branch-B",
  "status" : "running",
  "name" : "jenkins",
  "target_url" : "https://<jenkins-server>/jenkins/job/GitlabMergeRequests/679/",
  "description" : null,
  "created_at" : "2017-03-15T12:14:45.945Z",
  "started_at" : "2017-03-15T12:14:45.959Z",
  "finished_at" : null,
  "allow_failure" : false,
  "author" : {
    "name" : "Administrator",
    "username" : "root",
    "id" : 1,
    "state" : "active",
    "avatar_url" : "https://<gitlab-server>/uploads/user/avatar/1/avatar.png",
    "web_url" : "https://<gitlab-server>/root"
  }
}
```

### Problem description
For our project we have a single jenkins job setup to build all MR:s. The job has *Execute concurrent builds if necessary* enabled to allow multiple MR:s to be built in parallell. 

I recently noticed that some commits had double status objects in gitlab:

image

Curiously the two jobs point to different jenkins builds, each for a different MR. The running status is for the MR which contains the current commit, but the pending status belongs to a job that is building a completely different MR. The current pipeline for that MR looks like this:

image

Job 1204 har the same URL as job 1203 in the previous image.

Until now we've been spared from extra pending commit statuses since #489 was fixed. What has changed on our end is that we are doing many more builds than before and sometimes a build queue with waiting MR builds has formed. My hunch is that the wrong build number is used when the pending status is created. The code seems to use `job.getNextBuildNumber()`, which I guess doesn't work out when multiple builds are queued up.

ttsalonen:
Fixes #525 
Removed the build number from target url when the state is pending (job is waiting for next available executor).

Build number in the url causes problems when there are more than one job in the build queue. In those cases job.getNextBuildNumber returns the same number for every build in the queue which leads to wrong urls reported to Gitlab. Also, the link to job that is waiting in queue goes to 404 page anyway.

The best option for pending builds is probably just to set the target url point to the job page without the build number.

The problem is also described in issue https://github.com/jenkinsci/gitlab-plugin/issues/525

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/trigger/handler/AbstractWebHookTriggerHandler.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/trigger/handler/AbstractWebHookTriggerHandler.java
@@ -18,11 +18,11 @@ import hudson.model.Job;
 import hudson.plugins.git.GitSCM;
 import hudson.plugins.git.RevisionParameterAction;
 import hudson.scm.SCM;
-import jenkins.model.Jenkins;
 import jenkins.model.ParameterizedJobMixIn;
 import jenkins.triggers.SCMTriggerItem;
 import net.karneim.pojobuilder.GeneratePojoBuilder;
 import org.eclipse.jgit.transport.URIish;
+import org.jenkinsci.plugins.displayurlapi.DisplayURLProvider;
 
 import javax.ws.rs.ProcessingException;
 import javax.ws.rs.WebApplicationException;
@@ -69,8 +69,7 @@ public abstract class AbstractWebHookTriggerHandler<H extends WebHook> implement
                 if (client == null) {
                     LOGGER.log(Level.SEVERE, "No GitLab connection configured");
                 } else {
-                    String targetUrl =
-                        Jenkins.getInstance().getRootUrl() + job.getUrl() + job.getNextBuildNumber() + "/";
+                    String targetUrl = DisplayURLProvider.get().getJobURL(job);
                     client.changeBuildStatus(buildStatusUpdate.getProjectId(), buildStatusUpdate.getSha(),
                         BuildState.pending, buildStatusUpdate.getRef(), publisher.getName(), targetUrl, BuildState.pending.name());
                 }
--- a/src/main/java/com/dabsquared/gitlabjenkins/trigger/handler/push/OpenMergeRequestPushHookTriggerHandler.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/trigger/handler/push/OpenMergeRequestPushHookTriggerHandler.java
@@ -22,10 +22,10 @@ import hudson.model.CauseAction;
 import hudson.model.Job;
 import hudson.plugins.git.RevisionParameterAction;
 import hudson.triggers.Trigger;
-import jenkins.model.Jenkins;
 import jenkins.model.ParameterizedJobMixIn;
 import jenkins.model.ParameterizedJobMixIn.ParameterizedJob;
 import org.eclipse.jgit.transport.URIish;
+import org.jenkinsci.plugins.displayurlapi.DisplayURLProvider;
 
 import javax.ws.rs.ProcessingException;
 import javax.ws.rs.WebApplicationException;
@@ -158,7 +158,7 @@ class OpenMergeRequestPushHookTriggerHandler implements PushHookTriggerHandler {
                 (GitLabCommitStatusPublisher) ((AbstractProject) job).getPublishersList().get(GitLabCommitStatusPublisher.class);
             GitLabClient client = job.getProperty(GitLabConnectionProperty.class).getClient();
             try {
-                String targetUrl = Jenkins.getInstance().getRootUrl() + job.getUrl() + job.getNextBuildNumber() + "/";
+                String targetUrl = DisplayURLProvider.get().getJobURL(job);
                 client.changeBuildStatus(projectId, commit, BuildState.pending, ref, publisher.getName(), targetUrl, BuildState.pending.name());
             } catch (WebApplicationException | ProcessingException e) {
                 LOGGER.log(Level.SEVERE, "Failed to set build state to pending", e);


omehegan:
@ttsalonen thanks a lot for this PR! I have been busy for the last month but now am back to work on the plugin. I will look into reproducing this and verifying the fix ASAP.

ttsalonen:
@omehegan Have you been able to reproduce the issue? Please note that pull request https://github.com/jenkinsci/gitlab-plugin/pull/693 contains also these changes among others.

omehegan:
@ttsalonen I have been trying to reproduce the problem you described here, but I can't. Are you referring to Pipeline jobs or Freestyle? I created a sample Pipeline job which specifies a node label that does not exist. That job will stay in the queue forever.

```
updateGitlabCommitStatus name: 'build', state: 'pending'

node('foo') {
    sh 'echo "hi"'
}
```

When I trigger this with a GitLab merge request, it does get a specific build number even when it is in the queue, and the Pipeline/External job link in the GitLab UI points to that specific number. There's no 404. When I do something similar with a Freestyle job, it doesn't send anything to GitLab until the job starts, so I don't see how GitLab would see the pending status and generate any link to the Jenkins job. I see that #525 refers to Maven jobs. Is there something else I am missing in terms of how to reproduce this?

omehegan:
@ttsalonen I see in one sense my test may not have been valid, because I was using a hack to enforce the queue state. I guess if I tested with no executors available I might get a little bit different result. But I still don't understand how the "pending" status gets sent if the job has not actually started yet.

ttsalonen:
@omehegan The problem is with freestyle jobs (and also with Maven jobs I suppose). Sending 'Pending' status does not work with Pipelines currently (that feature it is added in https://github.com/jenkinsci/gitlab-plugin/pull/693).

You can reproduce the problem this way:

- Create a freestyle job that takes a while to execute (For example just execute shell "sleep 300") and prevent concurrent builds (or have just 1 executor)

- Create three branches in Gitlab with some commits and create merge requests from them

- Trigger build for branch1 from Gitlab and see that the build starts in Jenkins

- Trigger build for branch2 from Gitlab and see that the build goes into queue in Jenkins. If you follow the link in Gitlab for this 'Pending' build you should go to 404 page in Jenkins.

- Trigger build for branch3 from Gitlab and it also goes to queue. The link that the plugin sends to Gitlab for this pending build has a same build number as the link for previous branch2 build. The link goes to 404 page while the the branch2 and branch3 builds are in queue. After branch2 build starts the link goes to running build of branch2.

The plugin should send the 'Pending' status updates to Gitlab if you have configured the "Publish build status to Gitlab commit" post-build action. The implementation for that is in AbstractWebHookTriggerHandler and OpenMergeRequestPushHookTriggerHandler. The 'Pending' status is sent when the build is triggered. When the build is started, the plugin sends the 'Running' status update where links are correct.

omehegan:
@ttsalonen as far as I know, you CAN send pending status for Pipeline jobs. You just have to do it somewhat explicitly. See the doc here: https://github.com/jenkinsci/gitlab-plugin#scripted-pipeline-jobs  Does that not seem correct to you?

Thanks for your repro steps for the Freestyle job, that all makes sense. Let me just do some testing of the fix.

ttsalonen:
@omehegan Yes, it's true that pending status can be sent also in Pipeline jobs, but it can't be sent at the time when the job is triggered, only after the pipeline has already been started. Getting the pending status sent to Gitlab at the triggering time is a very nice feature if jobs have to wait in the build queue for a long time (sometimes several hours in my case).

omehegan:
@ttsalonen I see, that makes sense. Thanks for explaining! I'll work on testing this fix as soon as I can.

omehegan:
@ttsalonen I just merged #693 and it seems that created merge conflicts with this PR. Can you fix them? Or is this PR not needed now that #693 is merged?

ttsalonen:
@omehegan Yes, this is not needed anymore now that https://github.com/jenkinsci/gitlab-plugin/pull/693 is merged. I close this now.

PR #683, opened 2017-12-08
Make it possible to send pending build status to gitlab when a pipeline job is queued

ttsalonen:
It seems that it is currently not possible to update the build status to 'pending' in Gitlab immediately when a pipeline job is triggered. This is not good if pipeline has to wait a long time in a queue before the execution starts. Updating the status to 'pending' works for normal jobs.

This pull request contains one way to implement the 'pending' status update for pipelines but it is probably not the best possible. So, if someone who is more familiar with Gitlab-plugin and Jenkins in general has an idea how this could be implemented cleanly, I might have some time to work on it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -167,6 +167,11 @@
       <version>1.15</version>
     </dependency>
     <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-job</artifactId>
+      <version>1.15</version>
+    </dependency>
+    <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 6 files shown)


mreichel:
@ttsalonen since you closed the issue I guess I don't have to look at the code, right?

ttsalonen:
@mreichel I created another pull request that contains these changes and also a possibility to cancel pending builds on merge request updates: https://github.com/jenkinsci/gitlab-plugin/pull/693

Marked it as WIP because I haven't tested it yet properly but it would be great to get some comments about the implementation already (especially about how the 'pending build name' is configured for pipelines).

Issue #417, opened 2016-09-13
Create sample Jenkinsfiles to show users how to configure for Pipeline


omehegan:
We should include some sample Jenkinsfiles with the repo which walk users through how to use the plugin with Pipeline jobs.

Alien2150:
I can provide at least some working example with build / archive steps if needed

tonka3000:
Is there any commit so far which show these example files?

kadrach:
I would really appreciate an example, struggling to set this up properly.

Alien2150:
What excatly is needed?

JaniFuturemark:
Jenkins ver. 2.32.1
GitLab plugin ver. 1.4.4

Pipeline synxtax generator does not work so I am struggling when I am trying to add build trigger to jenkins file. An example would be awesome :)

```
properties([[$class:` 'GitLabConnectionProperty', gitLabConnection: 'Gitlab'], pipelineTriggers([<object of type com.dabsquared.gitlabjenkins.GitLabPushTrigger>])])
```

JaniFuturemark:
Got it. Correct line was:

```
properties([[$class: 'GitLabConnectionProperty', gitLabConnection: 'Gitlab'], pipelineTriggers([[$class: 'GitLabPushTrigger', branchFilterType: 'NameBasedFilter', triggerOnPush: true, includeBranchesSpec : 'MyBranchName', excludeBranchesSpec: '' ]])])
```

colourmeamused:
@JaniFuturemark #488 will make the syntax friendlier. Also included a basic example of a Jenkinsfile using the declarative syntax, more to follow!

You still need to take #475 into account though - you need to define all job properties as code the properties step.

jgeorgeson:
I would like to see this coupled with a plain API-like documentation of available pipeline methods/objects (both scripted and declarative).

omehegan:
@jgeorgeson what would be an example of "API-like" documentation? All these doc issues are something I plan to address soon.

jgeorgeson:
[...]

PR #684, opened 2017-12-10
Improve Jenkinsfile sample code in README.md

guessi:
Currently, there's no documentation for how to configure [Scripted Jenkinsfile](https://jenkins.io/doc/book/pipeline/syntax/#scripted-pipeline), it would be better to have sample code in README.md

Related: #417

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,9 @@ Due to this the plugin just listens for GitLab Push Hooks for multibranch pipeli
 
 Example `Jenkinsfile` for multibranch pipeline jobs
 ```
+// Define your secret project token here
+def project_token = 'abcdefghijklmnopqrstuvwxyz0123456789ABCDEF'
+
 // Reference the GitLab connection name from your Jenkins Global configuration (http://JENKINS_URL/configure, GitLab section)
 properties([
     gitLabConnection('your-gitlab-connection-name'),
--- a/README.md
+++ b/README.md
@@ -143,7 +143,19 @@ Due to this the plugin just listens for GitLab Push Hooks for multibranch pipeli
 Example `Jenkinsfile` for multibranch pipeline jobs
 ```
 // Reference the GitLab connection name from your Jenkins Global configuration (http://JENKINS_URL/configure, GitLab section)
-properties([gitLabConnection('your-gitlab-connection-name')])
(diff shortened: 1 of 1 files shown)


omehegan:
@guessi thanks for improving this! But I think your diff applies to all scripted Pipeline, not just Multibranch, so I would like to refactor the README further to make that clearer. If you want to do that, feel free, otherwise I will work this into a larger rewrite of this part of the doc.

guessi:
@omehegan good to know it's going to be refactor, but I'm not sure what do you mean by refactor, if it is to rewrite or refactor document structure, I'm afraid I have no time for that ... :'(

omehegan:
@guessi I just mean that your commit edits the part of the README that is specific to Multibranch project. I believe your change applies to any type of Pipeline project, so I want to make sure that is clear to the reader. I will modify it :)

guessi:
Rebase onto current master (80c3786) for fixing merge conflict

omehegan:
@guessi I have just done a major refactor of the README, I think it's must easier to follow now. I specifically reorganized the documentation around Pipeline.

I have looked at your PR again, and I'm a little confused as to whether it is valid or not. Here's why: if the job is a regular, non-multibranch Pipeline, then all of this configuration is exposed in the Jenkins UI. There is no reason you would need to store it in the Jenkinsfile. On the other hand, if the job is a Multibranch type, AFAIK the plugin does not parse the JSON request from GitLab, so none of the data would be read. And thus these settings like ciSkip and possibly even secretToken would be ignored. I need to test this, but I was wondering if you think I've misunderstood something here. Please let me know!

guessi:
@omehegan the original purpose of this PR is to resolve missing example for `Scrippted Pipeline`, the regular Jenkinsfile, and I've checked the sample code in this PR again, there should have NO credential to be shown on the Jenkins WebUI, and I've tested that with/without `secretToken`, `ciSkip` are worked.

omehegan:
@guessi sorry for all the back-and-forth, it had been a few months since I looked at this code and I got myself confused. Your change is correct, I expanded on it and put it in a different part of the README. Thank you!

guessi:
NP 👍

PR #686, opened 2017-12-14
Improve doc to setup webhook with folders

smutel:
Add a comment to improve accuracy when setup a webhook with Jenkins Multibranch Pipeline/Workflow job inside folders.


omehegan:
@smutel thanks, doc improvements are always welcome!

PR #698, opened 2018-01-02
Make plugin respect no-proxy-hosts when a proxy is used

### LINKED ISSUE #697: Gitlab plugin ignores NO_PROXY field of HTTP proxy settings
My Jenkins setup operates with set of intranet and internet resources. Internet access requires proxy settings. I added proxy configuration to Jenkins settings and also add all local resources to NO_PROXY list. With the configuration Gitlab plugin reports HTTP 401 error on each HTTP request to local Gitlab. Plugin works correctly, If I totally disable proxy.

The issue is localized at https://github.com/jenkinsci/gitlab-plugin/blob/316b8bebb8927081547d31333ec76e1c97b98f25/src/main/java/com/dabsquared/gitlabjenkins/gitlab/api/impl/ResteasyGitLabClientBuilder.java#L98

But it looks like RestEasyClientBuilder does not support NO_RPOXY. I'm ready to contribute fix, if you have any hint how to implement NO_PROXY logic.

Argelbargel:
Fixes #697 
This should fix #697 ensuring that the ResteasyClient-Builder respects non-proxy-hosts in setups using a proxy to connect to the internet

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/gitlab/api/impl/ResteasyGitLabClientBuilder.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/gitlab/api/impl/ResteasyGitLabClientBuilder.java
@@ -53,7 +53,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-import static java.net.Proxy.NO_PROXY;
+import static java.net.Proxy.Type.HTTP;
 
 
 @Restricted(NoExternalUse.class)
@@ -90,19 +90,22 @@ public class ResteasyGitLabClientBuilder extends GitLabClientBuilder {
 
     private GitLabClient buildClient(String url, String apiToken, ProxyConfiguration httpProxyConfig, boolean ignoreCertificateErrors, int connectionTimeout, int readTimeout) {
         ResteasyClientBuilder builder = new ResteasyClientBuilder();
+
         if (ignoreCertificateErrors) {
             builder.hostnameVerification(ResteasyClientBuilder.HostnameVerificationPolicy.ANY);
             builder.disableTrustManager();
         }
 
-        Proxy proxy = httpProxyConfig != null ? httpProxyConfig.createProxy(getHost(url)) : NO_PROXY;
-        if (proxy != NO_PROXY) {
-            InetSocketAddress address = (InetSocketAddress) proxy.address();
-            builder.defaultProxy(address.getHostName().replaceFirst("^.*://", ""),
-                address.getPort(),
-                address.getHostName().startsWith("https") ? "https" : "http",
-                httpProxyConfig.getUserName(),
-                httpProxyConfig.getPassword());
+        if (httpProxyConfig != null) {
+            Proxy proxy = httpProxyConfig.createProxy(getHost(url));
+            if (proxy.type() == HTTP) {
+                InetSocketAddress address = (InetSocketAddress) proxy.address();
+                builder.defaultProxy(address.getHostString().replaceFirst("^.*://", ""),
+                    address.getPort(),
+                    address.getHostName().startsWith("https") ? "https" : "http",
+                    httpProxyConfig.getUserName(),
+                    httpProxyConfig.getPassword());
+            }
         }
 
         GitLabApiProxy apiProxy = builder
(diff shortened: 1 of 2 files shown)


src/main/java/com/dabsquared/gitlabjenkins/gitlab/api/impl/ResteasyGitLabClientBuilder.java
```diff
@@ -90,19 +90,22 @@ public final GitLabClient buildClient(String url, String apiToken, boolean ignor
...
     private GitLabClient buildClient(String url, String apiToken, ProxyConfiguration httpProxyConfig, boolean ignoreCertificateErrors, int connectionTimeout, int readTimeout) {
         ResteasyClientBuilder builder = new ResteasyClientBuilder();
+
         if (ignoreCertificateErrors) {
             builder.hostnameVerification(ResteasyClientBuilder.HostnameVerificationPolicy.ANY);
             builder.disableTrustManager();
         }
 
-        Proxy proxy = httpProxyConfig != null ? httpProxyConfig.createProxy(getHost(url)) : NO_PROXY;
-        if (proxy != NO_PROXY) {
```
Argelbargel: This change ignores socks-proxy-setups as we currently force the proxy's protocol to http(s) anyway. Does anyone know whether the plugin ever worked with socks-proxys?

src/main/java/com/dabsquared/gitlabjenkins/gitlab/api/impl/ResteasyGitLabClientBuilder.java
```diff
@@ -90,19 +90,22 @@ public final GitLabClient buildClient(String url, String apiToken, boolean ignor
...
-            InetSocketAddress address = (InetSocketAddress) proxy.address();
-            builder.defaultProxy(address.getHostName().replaceFirst("^.*://", ""),
-                address.getPort(),
-                address.getHostName().startsWith("https") ? "https" : "http",
-                httpProxyConfig.getUserName(),
-                httpProxyConfig.getPassword());
+        if (httpProxyConfig != null) {
+            Proxy proxy = httpProxyConfig.createProxy(getHost(url));
+            if (proxy.type() == HTTP) {
```
Argelbargel, line 101: This change ignores socks-proxy-setups as we currently force the proxy's protocol to http(s) anyway. Does anyone know whether the plugin ever worked with socks-proxys?

mreichel (approved):

src/main/java/com/dabsquared/gitlabjenkins/gitlab/api/impl/ResteasyGitLabClientBuilder.java
mreichel: I don't know if it ever did.

omehegan: Me either.

omehegan:
I want to see if we get some more feedback in the original issue before I merge this.

src/main/java/com/dabsquared/gitlabjenkins/gitlab/api/impl/ResteasyGitLabClientBuilder.java
```diff
@@ -53,7 +53,7 @@
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-import static java.net.Proxy.NO_PROXY;
+import static java.net.Proxy.Type.HTTP;
```
christ66, line 56: Is it possible to instead rely on `java.net.Proxy.*` we could use the [Async Http Client Plugin](https://wiki.jenkins.io/display/JENKINS/Async+Http+Client+Plugin) (which IIRC has the proxy configuration in the UI under Manage Jenkins > Manage Plugins > Advanced tab)?

Argelbargel: The UI unter Manage Plugins is the one built into Jenkins itself. ResteasyGitLabClientBuilder uses them already. What i could do is to use Proxy.Type.DIRECT instead of Proxy.Type.HTTP in the check to allow SOCKS-proxied connections but then we'd have to create a socks-configuration further down. As of relying on the Async Http Client Plugin - i'm not sure whether it is possible to use this with Resteasy

omehegan:
Still waiting for feedback from the original issue submitter.

omehegan:
This was just released in version 1.5.5 of the plugin. It should be available in the update center soon.

naxim92:
Version 1.7.15 still no changes. Still ignoring NO_PROXY. See https://github.com/jenkinsci/gitlab-plugin/issues/1562

PR #709, opened 2018-01-30
WIP: Filtering by label never matches anything

mreichel:
Filtering by label never matches anything #647 
Create a label object with a title instead of a name.


omehegan:
Superseded by #732.

Issue #181, opened 2016-01-15
Remove source branch with "Accept merge request on success"


sebastienbonami:
Hi, it would be nice to have the option to remove source branch automatically when the functionality "Accept merge request on success" is used in a Jenkins job.

Thanks!

omorillo:
I'm on it.

omorillo:
My implementation for this issue is now ready. 

Unfortunately, I observed a bug in the **GitLab API** implementation which results in the source branch always to be removed when using the _"Accept merge request on success"_ feature, no matter whether the **"should_remove_source_branch"** flag was enabled or not. I could observe this behaviour in **GitLab CE** version **8.7.3** and **8.7.6**.

Creating an issue on the **GitLab bug tracker** (https://gitlab.com/gitlab-org/gitlab-ce/issues/17992).

As the feature can not properly work at the moment, the changeset will not be added to the master yet. A pull request will be put in place which will be merged to master, once the issue on **GitLab** site is fixed.

sebastienbonami:
Well... this feature is now present on GitLab's side so I don't need this anymore.

PR #735, opened 2018-04-02
Pipeline options

ptzafrir:
Provide two configuration options for pipeline users when using acceptGitlabMergeRequest step

- useMRDescription, which will build   a commit message similar to gitlab "include description in merge commit" option.
- removeSourceBranch to delete source branch after accepting the Mr

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/workflow/AcceptGitLabMergeRequestStep.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/workflow/AcceptGitLabMergeRequestStep.java
@@ -35,9 +35,12 @@ public class AcceptGitLabMergeRequestStep extends Step {
 
     private String mergeCommitMessage;
 
+	private boolean useMRDescription;
+
     @DataBoundConstructor
-    public AcceptGitLabMergeRequestStep(String mergeCommitMessage) {
+    public AcceptGitLabMergeRequestStep(String mergeCommitMessage,boolean useMRDescription) {
         this.mergeCommitMessage = StringUtils.isEmpty(mergeCommitMessage) ? null : mergeCommitMessage;
+        this.useMRDescription = useMRDescription;
     }
 
 	@Override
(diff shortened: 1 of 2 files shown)


omehegan:
Thanks for the PR, it sounds good! I'll try to get it tested and into the next release.

src/main/java/com/dabsquared/gitlabjenkins/workflow/AcceptGitLabMergeRequestStep.java
```diff
@@ -89,7 +105,19 @@ protected Void run() throws Exception {
             return null;
         }
 
-        private void println(String message) {
+        private String getCommitMessage(MergeRequest mergeRequest) {
+        	if (!step.useMRDescription)
+        		return step.mergeCommitMessage;
+        	
+        	println("Merge description is: "+mergeRequest.getDescription());
```
mattiasb: Should this be using the `LOGGER` object instead?

ptzafrir: This is calling the println method defined below, which will use LOGGER, or send output to listener if there is a task listener in use.

ptzafrir: Looking at it again, it seems like a debug comment that I wouldn't like to include (will output needless data to the job console). I removed it. Thanks for catching this.

mattiasb: Oh ok! :) Didn't see that one.

omehegan:
@mattiasb thanks a lot for reviewing this!

mattiasb:
@omehegan Oh that was just a drive-by comment. GitHub makes this up to be more than it was. :)

omehegan:
@mattiasb it's more review than I did :)

omehegan:
@ptzafrir I would like to get this merged for release soon, but I have two requests. First, can you add a note in the README which explains the new options? Second, I see you are adding a "remove source branch" option. We had an open feature request for this, which was closed recently because GitLab has added this feature on their side. See #181. Since your feature for that would only apply to Pipeline jobs, maybe we should just leave it out?

ptzafrir:
I've updated the README.md. As for the option to remove source branch, in my use case I work with a team with a strict policy to remove all feature branches following merge. Some team members forget to manually check the checkbox when creating MR and so we are left to manually cleanup every once in a while.

omehegan:
@ptzafrir thanks! I will merge this now, but I just released a few other things earlier today so I will hold this for the next release. Should be in a couple of weeks.

omehegan:
Released as 1.5.8 today because we had an unrelated breaking change in 1.5.7.

PR #760, opened 2018-05-03
Readme: escape markdown refspec for tags

verath:
The `*` in the refspec was instead interpreted as markdown cursive. Surrounded the refspec in markdown inline code block to prevent this.


omehegan:
Good catch, thank you!

PR #767, opened 2018-05-18
Fix NPE in GitLabPushTrigger

MateuszJajeGP:
because missing of it was causing NPE during processing pipeline hook


omehegan:
This fix was released in version 1.5.6 of the plugin.

MateuszJajeGP:
Great! Thanks

Issue #785, opened 2018-06-25
Defined Parameters not set/empty


azerbe:
### Context
- **Gitlab plugin version**: 1.5.6
- **Gitlab version**: 11.0.0-ee
- **Jenkins version**: 2.121.1 (jenkinsci/blueocean docker)
- **Job type**: Freestyle
- **Environment Injector Plugin version**: 2.1.5

### Logs & Traces
**injectedEnvVars.txt** (from EnvInject Plugin)
```
gitlabActionType=MERGE
gitlabBranch=mergerequest
gitlabMergeRequestAssignee=Z
gitlabMergeRequestId=5
gitlabMergeRequestIid=5
gitlabMergeRequestLastCommit=712579185a9b572ec7cad6d52c00f90e49205663
gitlabMergeRequestState=opened
gitlabMergeRequestTargetProjectId=21
gitlabMergeRequestTitle=Mergerequest Test
gitlabMergedByUser=Z
gitlabSourceBranch=
gitlabSourceNamespace=Z
gitlabSourceRepoHomepage=https://git@....../Z/jenkins-docker-maven-example
gitlabSourceRepoHttpUrl=https://git@....../Z/jenkins-docker-maven-example.git
gitlabSourceRepoName=jenkins-docker-maven-example
gitlabSourceRepoSshUrl=ssh://git@....../Z/jenkins-docker-maven-example.git
gitlabSourceRepoURL=ssh://git@....../Z/jenkins-docker-maven-example.git
gitlabTargetBranch=
gitlabTargetNamespace=Z
gitlabTargetRepoHttpUrl=https://git.aqube.de/Z/jenkins-docker-maven-example.git
gitlabTargetRepoName=jenkins-docker-maven-example
gitlabTargetRepoSshUrl=ssh://git@....../Z/jenkins-docker-maven-example.git
gitlabUserEmail=mymail
gitlabUserName=myname
```

**JenkinsJob config.xml**
```
<?xml version='1.1' encoding='UTF-8'?>
<project>
  <actions/>
  <description>Test Project to test Jenkins&lt;-&gt;GitLab integration</description>
  <keepDependencies>false</keepDependencies>
  <properties>
    <com.dabsquared.gitlabjenkins.connection.GitLabConnectionProperty plugin="gitlab-plugin@1.5.6">
      <gitLabConnection>GitLab</gitLabConnection>
    </com.dabsquared.gitlabjenkins.connection.GitLabConnectionProperty>
[...]

PR #791, opened 2018-06-29
Update Readme to fix param example for #356 & #785

### LINKED ISSUE #356: "Parametrized builds" section of README is out of date
### Context
- **Gitlab plugin version**: 1.2.3
- **Gitlab version**: 8.8.3
- **Jenkins version**: 2.8
### Problem description

I had some trouble upgrading to latest versions of everything. Specifically with variables/parameters being passed to the plugin. I understood that instead of parameters they now got passed as env variables, but it looked like some of them where empty. I did read the linked pages on jenkins security update and EnvInject plugin but it wasn't clear what's going on.

I finally figured out that you have to REMOVE the parameters. and make your merge-request job be un-parametrized. Any parameter specified (E.g. gitlabTargetBranch) caused the corresponding env variable to NOT get populated. Remove all parameters fixed the problem.

So my only complaint that this wasn't clear from README and that "Parametrized builds" section still suggests making parameters which won't work now with latest jenkins/plugin

ElijahLynn:
The commit for https://github.com/jenkinsci/gitlab-plugin/commit/aa8c1011b88d5b33c0f689b6ce257257b58acfaa was off a little. This should fix #356 & #785.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ if (env['gitlabTargetBranch'] != null) {
 return map
 ```
 
-You will need to update this code anytime you add or remove parameters.
+You can then reference these variables in your job config, e.g. as `${sourceBranch}`. You will need to update this code anytime you add or remove parameters.
 
 ## Git configuration 
 ### Freestyle jobs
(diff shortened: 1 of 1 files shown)


omehegan:
@ElijahLynn thanks a lot for clarifying this!

PR #798, opened 2018-07-09
Skip build triggering for approved action when checking state changes

### LINKED ISSUE #796: Build triggered on MR approval even if option is disabled
## Issue

I upgraded the plugin to 1.5.8 and this option is broken. I have this option unchecked, but the build is still triggered upon an approval.

This option was broken a long time ago, then it was fixed, then broken again in the last few updates (I think either 1.5.7 or 1.5.8)

### Context
- **Gitlab plugin version**: 1.5.8
- **Gitlab version**: 11.0.0
- **Jenkins version**: 2.130
- **Job type**: Freestyle

keijohyttinen:
Fixes #796

In plugin's `MergeRequestHookTriggerHandlerImpl` implementation, State "updated" is enabled by "_Rebuild open Merge Requests_" option in GUI.

As gitlab EE offers option to trigger approved MRs, this change guarantees that when "approved" Action is received altogether e.g. with State "updated" attribute, it won't trigger Jenkins build unless "approved" option is enabled for the job. 

Improved test cases to cover more use cases including this scenario.

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/trigger/handler/merge/MergeRequestHookTriggerHandlerImpl.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/trigger/handler/merge/MergeRequestHookTriggerHandlerImpl.java
@@ -42,6 +42,7 @@ class MergeRequestHookTriggerHandlerImpl extends AbstractWebHookTriggerHandler<M
     private final boolean skipWorkInProgressMergeRequest;
 	private final Collection<Action> allowedActions;
     private final EnumSet<Action> skipBuiltYetCheckActions = EnumSet.of(Action.open, Action.approved);
+    private final EnumSet<Action> skipAllowedStateForActions = EnumSet.of(Action.approved);
     private final boolean cancelPendingBuildsOnUpdate;
 
     MergeRequestHookTriggerHandlerImpl(Collection<State> allowedStates, boolean skipWorkInProgressMergeRequest, boolean cancelPendingBuildsOnUpdate) {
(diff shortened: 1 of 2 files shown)


keijohyttinen:
@chinhodado Unfortunately I don't have gitlab EE to verify whether this fix works in real environment with gitlab EE. Could you please test this fix in advance with your setup?

chinhodado:
@keijohyttinen Thank you for the fix. I will test this later, but one thing I just noticed is, it seems an approval only triggers the build if **it's the last approval required**. So for example, if the merge request is set to require 2 approvals, then the first approval will not trigger the build, but the second will. Does this sound related to this fix?

keijohyttinen:
@chinhodado As far as I understand, that's not related to this fix but believe it's a feature in Gitlab side. Sounds like approved action hook won't be triggered by Gitlab until last approval given. But please notice that I can't be sure since cannot verify it or I've missed something in this plugin's side...

chinhodado:
Ok I have tested the change and here is the result:

 - With "Approved Merge Requests (EE-only)" unchecked, build is never triggered on approval. **This fixes my issue, so that's great!**

- With "Approved Merge Requests (EE-only)" unchecked, **build is only triggered on the last required approval** (e.g. if MR requires 2 approvals then build is only triggered on the second approval). I think this is also correct behavior.

jbobaggins:
Any idea on when this will make it into the next release?  We've hit this problem and its really annoying.  

Thanks for fixing and testing @chinhodado and @keijohyttinen !

omehegan:
I built a snapshot release of this, it's available at http://repo.jenkins-ci.org/snapshots/org/jenkins-ci/plugins/gitlab-plugin/1.5.9-SNAPSHOT/gitlab-plugin-1.5.9-20180717.064808-1.hpi  I would like to confirm the fix with others who were affected before I merge it, because I don't have a GitLab EE install either. @jbobaggins can you try this ^

omehegan:
This was released in version 1.5.9 of the plugin.

PR #799, opened 2018-07-10
gitlabCommitStatus step should process name arg as optional

### LINKED ISSUE #795: gitlabCommitStatus missing required parameter
Might be related to #789
Seems like 1.5.8 is still broken for us around `gitlabCommitStatus`. We use it as described in the docs as:
```
 gitlabCommitStatus {
       // The result of steps within this block is what will be sent to GitLab 
       sh 'mvn install'
    }
```
And are getting:
```
Missing required parameter: "name" @ line 104, column 21.
                       gitlabCommitStatus{
```
Reverting back to 1.5.6 resolved the issue.

stchar:
Fixes  #795 
Name argument should be optional
Captured on 
Jenkins: 2.121.1
Gitlab: 10.8.2-ee
Gitlab plugin 1.5.8


omehegan:
This was released in version 1.5.9 of the plugin.

PR #801, opened 2018-07-16
Expose cause information to API queries

mkobit:
…I queries

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/cause/CauseData.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/cause/CauseData.java
@@ -8,6 +8,8 @@ import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.builder.EqualsBuilder;
 import org.apache.commons.lang.builder.HashCodeBuilder;
 import org.apache.commons.lang.builder.ToStringBuilder;
+import org.kohsuke.stapler.export.Exported;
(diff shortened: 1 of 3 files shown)


omehegan:
Thanks for this PR! Can you elaborate how you would see it being used?

mkobit:
Hey @omehegan !

We have been running into flaky builds and we would like to be able to use the API to collect some data around those failures. We are hoping to avoid having to use admin APIs (like the Groovy script console) so that anybody on our Jenkins instance can use their API token for builds they have access to to get this information.

I don't know if this is the best route to take, but seemed like a simple way to go.

There might need to be some tweaking of http://stapler.kohsuke.org/apidocs/org/kohsuke/stapler/export/Exported.html#visibility-- to make them more visible in certain circumstances, but this seemed like an erasy route to go down for exposing the information. For example, a few other `Cause` types that export information to the API that might be used as references:

* http://javadoc.jenkins-ci.org/hudson/model/Cause.RemoteCause.html
* http://javadoc.jenkins-ci.org/hudson/model/Cause.UserIdCause.html
* http://javadoc.jenkins-ci.org/hudson/model/Cause.UpstreamCause.html

omehegan:
@abayer could you possibly comment on whether you think this is a good approach or not? I just have not ever tried to expose anything to the Jenkins API before.

abayer (approved):
Seems reasonable to me.

PR #806, opened 2018-07-17
Update README.md to point to new location for gitlab-plugin CI

padyx:
I noticed that the link in the README.md file points to the old location for the Jenkins Cloudbees location where the job to test PRs is disabled.

This pull request only updates the README file to point to the new location which I've determined from the checks of recent PRs.

Old: https://jenkins.ci.cloudbees.com/job/plugins/job/gitlab-plugin/
New: https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fgitlab-plugin/


omehegan:
Good catch, thanks!

Issue #794, opened 2018-07-05
NullPointerException in Trigger for "Build on successful pipeline events"


Persi:
## Issue
NullPointerException on using Gitlab pipeline events.

### Context
- **Gitlab plugin version**: 1.5.8
- **Gitlab version**: 10.2.5
- **Jenkins version**: 2.129
- **Job type**: Pipeline

### Gitlab hook sent data
[JSON Data](https://pastebin.com/RsDTcAyT)

### Occured exception:
[Stacktrace](https://pastebin.com/bcLy7xnU)

### Problem description

I am using a gitlab-ci Pipeline als CI Pipeline for my project. To deploy the artifact, created by the gitlab-ci pipeline, on a system in another environment, I need to use a Jenkins pipeline. So I tried to configure my Jenkins job for Gitlab hooks, by using the Gitlab plugin. If I send an event from Gitlab to Jenkins on any push, the Jenkins pipeline is started correctly. But I need to start my Jenkins pipeline only after the gitlab-ci pipeline succeeded, so I only send Pipeline Events from Gitlab. I have configured in the Gitlab plugin configuration for my pipeline project, to allow all branches to trigger the build.

The pipeline events from Gitlab cause a NullPointerException in Jenkins, so the Jenkins pipeline is not triggered.

omehegan:
The NPE occurs here:
https://github.com/jenkinsci/gitlab-plugin/blob/master/src/main/java/com/dabsquared/gitlabjenkins/trigger/handler/pipeline/PipelineHookTriggerHandlerImpl.java#L66

Is it possible there is an issue with how you have configured the branch filtering? You said you have set it to allow all branches. Can you show this configuration?

Persi:
Hi @omehegan,

I did a short look into the sources as well, that is why I mentioned the branch filtering config.
I tried different filtering settings, from filtering by branchname, as well as regex and at least allow all. The NPE occured with every setting.

The config looks like following: [XML](https://pastebin.com/FUnAY4dh)
In the job configuration it looks like the following: [Screen](https://pasteboard.co/Ht9HHv4.png)

omehegan:
I will have to try and find time to reproduce this. Can you follow the instructions in the User Support section of the README, to capture some additional debug logs and attach them after reproducing the problem?

Persi:
[...]

PR #818, opened 2018-07-30
NPE-Fix for Bug 794: Initialize BranchFilter on PipelineHooks as well.

Persi:
Fix for Issue: https://github.com/jenkinsci/gitlab-plugin/issues/794

Reproduced mentioned Bug with the smallest possible setup.
Jobconfiguration looked like the following: [XML](https://pastebin.com/zLEFpefV)

Posting the GitLab hook data given in the above Issue caused NPE without the fix. After the fix NPE was gone.


omehegan:
Thanks! I'll get this released soon.

omehegan:

PR #754, opened 2018-04-18
Add option to filter source branch with regex

sebastienbonami:
Fixes #751

It works, but I didn't adjust any test classes for that. Can somebody do it? Thanks.

PR #824, opened 2018-08-13
Fix the tests broken in #754.

### LINKED ISSUE #751: Source Branch Regex
With this plugin, we can filter branches by regex using the "Target Branch Regex" option, but there's no "Source Branch Regex" option. Is there a special reason for that?

I'm currently looking at the code and I will try to implement a "Source Branch Regex" option since I need it for my use case.

bonzani:
Fix the filter tests for #754 that do not work anymore because of the sourceBranch argument in isBranchAllowed.

Fixes #751.


omehegan:
@bonzani thanks a lot for fixing the tests!

omehegan:

PR #826, opened 2018-08-23
Add support for System Hooks

### LINKED ISSUE #836: Support GitLab System Hooks
### Context
- **Gitlab plugin version**: 1.5.9
- **Gitlab version**: 10.7.3
- **Jenkins version**: 2.73.3
- **Job type**: Pipeline

### Logs & Traces
```
<html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=utf-8"/>
<title>Error 404 Not Found</title>
</head>
<body><h2>HTTP ERROR 404</h2>
<p>Problem accessing /project/CI-Code-Analysis. Reason:
<pre>    Not Found</pre></p><hr><a href="http://eclipse.org/jetty">Powered by Jetty:// 9.4.z-SNAPSHOT</a><hr/>

</body>
</html>
```

### Problem description
I add a global hook in gitlab ,but it return 404 when it trigger,but it works good when I add it in single project.I check the param,when it is global ,the head params like this:
```
Content-Type: application/json
X-Gitlab-Event: System Hook
```
But when it in project,the head params like this:
```
Content-Type: application/json
X-Gitlab-Event: Merge Request Hook
```

I look the code maybe about it at https://github.com/jenkinsci/gitlab-plugin/blob/518063be653de672c574cc86536c7617fb0e5159/src/main/java/com/dabsquared/gitlabjenkins/webhook/ActionResolver.java#L107
```
switch (eventHeader) {
            case "Merge Request Hook":
                return new MergeRequestBuildAction(project, getRequestBody(request), tokenHeader);
            case "Push Hook":
            case "Tag Push Hook":
                return new PushBuildAction(project, getRequestBody(request), tokenHeader);
            case "Note Hook":
                return new NoteBuildAction(project, getRequestBody(request), tokenHeader);
            case "Pipeline Hook":
                return new PipelineBuildAction(project, getRequestBody(request), tokenHeader);
            default:
                LOGGER.log(Level.FINE, "Unsupported X-Gitlab-Event header: {0}", eventHeader);
                return new NoopAction();
        }
```
So,the plugin do not handle the `System Hook` head.

How can I resolve the problem?Thanks

elliottmb:
Gitlab EE has a concept of System Hooks, which nearly match repo specific hooks. These allow you to organizationally trigger external event responses across all your projects, such as running security scanning tooling and license scanning tooling on change and provide feedback to MR's.

See https://docs.gitlab.com/ee/system_hooks/system_hooks.html

I've added basic support for the similar Push, Tag, and Merge Request event variants as they nearly map 1 to 1 with the repo specific ones. Included are tests and example json payloads based off a working Gitlab instance.

Fixes #836.


elliottmb:
@omehegan Let me know if any changes need to be made to this PR.

I am not a huge fan of the fact we have to essentially double parse the json body to add this support. I didn't want to go and introduce alternative constructors or change the constructors to take a Json Tree given the ObjectMapper usage already in place.

omehegan:
@elliottmb thanks for contributing this! Are you already using this patch on your Jenkins instance? This is probably an ignorant question, but why exactly did you not want to change the existing constructors? I guess it complicates things in a different way because the regular webhooks do not send a JSON tree?

elliottmb:
We've tested it, but I felt I wanted to see about mainlining the idea rather than maintaining a fork for some time.

As for the constructors, with the normal project level web hooks the header maps one to one (or pretty closely, with tags and pushes overlapping) so you can use that to then deserialize the payload directly into a matching data class. In the case of system hooks, they all have the same header, regardless of payload. So you essentially, as I've done in my changes, have to double look / deserialization. This seems potentially wasteful on a busy system, but may be an over optimization. The thought I had was adding additional constructors to take in a generic json tree from this early look and use that to build the data class directly. 

Again, that may be an over optimization, but the thought of parsing the json twice per system hook seemed wrong.

omehegan:
@elliottmb thanks for explaining! I'm contemplating what I think a "busy system" could look like. These System Hooks change the calculus a bit, if it means that potentially *every push to GitLab* is going to send data to Jenkins. Unless you *want* to work on a more optimized solution, I think I am OK to accept this as-is. It would be helpful if you added a comment to the code which specifically says that this is going to introduce additional parsing. That way if it does become a drag on performance in the future, it will be easy to find.

elliottmb:
@omehegan After looking at it with fresh eyes, I saw an easy way to avoid the double deserialization. So I've updated the PR with that change. Should be good to go now, unless you see something obvious. Thanks!

omehegan:
@elliottmb excellent, thanks! I will get this tested and merged soon. 

@pdudits if you have a few minutes and would be willing to review this change, I would appreciate it, but I don't think it's too critical.

elliottmb:
@omehegan @pdudits Any update on the review?

omehegan:
@elliottmb have not had time to do more with the plugin lately, but I will be at Jenkins World next week and should be able to focus on it then.

omehegan:
@elliottmb I am going to release this today. Can you also add something to the README about this? Maybe a new heading under https://github.com/jenkinsci/gitlab-plugin#job-trigger-configuration

omehegan:

elliottmb:
Nice! Sorry for not getting around to opening another pull request for the readme. I'll try to find time for that this week.

omehegan:
@elliottmb sounds good, you can make the PR close #840. I thought I could assign that to you, but I guess not because you're not a project member.

PR #828, opened 2018-08-25
Fix pipeline job mergeStrategy

Minipada:
Using the latest gitlab plugin version, I ran into the error described [here](https://issues.jenkins-ci.org/browse/JENKINS-51638).
Changing the mergeStrategy to DEFAULT made it work again.


omehegan:
Thanks!

PR #830, opened 2018-08-30
Update jenkins.version to 2.7.1 and test-harness to 2.4.0

deepy:
Update `jenkins.version` and `jenkins-test-harness.version` with the goal of enabling testing of pipeline steps.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,9 +7,9 @@
   </parent>
 
   <properties>
-    <jenkins.version>1.609.3</jenkins.version>
+    <jenkins.version>2.7.1</jenkins.version>
     <hpi-plugin.version>1.115</hpi-plugin.version>
-    <jenkins-test-harness.version>${jenkins.version}</jenkins-test-harness.version>
+    <jenkins-test-harness.version>2.40</jenkins-test-harness.version>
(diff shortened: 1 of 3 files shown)


Wadeck (changes requested):
There is a test failure, I imagine due to the change of Jetty implementation.
Otherwise always nice to see an update of plugin 👍

deepy:
@Wadeck help would be greatly appreciated as I currently got limited time and had to take a pause from this after failing to figure out why it fails :-)

omehegan:
@deepy any chance you will have time to look at this soon? Did you need help from @Wadeck?

Wadeck:
@omehegan @deepy I have investigated a bit. The root cause is the incompatibility between jenkins-test-harness and parent pom versions. Parent pom includes as dependency servlet-api:2.4 but in jenkins-test-harness, we expect 3.1+.

Quick advice: upgrade parent pom + jenkins-test-harness to latest version, that will solve *that* problem but perhaps add new ones. After that the plugin will be more up-to-date.

deepy:
Upgrading parent pom + jth to latest version resulted in some dependency juggling, will pursue that a bit further

omehegan:
@deepy do you think you will be able to work on this anytime soon?

Issue #631, opened 2017-09-25
Allow for Jenkins project/folder level GitLab Token API credentials to be used for GitLab connection


acohen716:
## Feature Request
In our enterprise we have an Enterprise GitLab and an Enterprise Jenkins set up such that groups have their own Jenkins folders and don't have access to the Global Jenkins configuration to add their GitLab API Token credentials for their GitLab groups projects at the global level. It appears that the gitlab-plugin pulls the GitLab connection info from the Manage Jenkins->Configure admin level which only the Enterprise Jenkins admin has access to. If the GitLab API Token was pulled from the project/folder/group level vs. the admin level then non-admins would be able to add their own GitLab API Token without requesting the Jenkins admin to add it at the global level.

### Context
- **Gitlab plugin version**: 1.4.2
- **Gitlab version**: Community Edition 9.5.1
- **Jenkins version**: [CloudBees Jenkins Enterprise 2.46.24.0.2-fixed](https://release-notes.cloudbees.com/release/1/2.46.24.0.2)
- **Job type**: Freestyle

bbodenmiller:
We're also in need of this. @omehegan anything we can do to help advance this idea?

omehegan:
@bbodenmiller I would welcome a pull request that implements it :)  Otherwise I'm not personally in a position to do it, I'm really just a shepherd for this plugin. Other volunteers write the code.

bbodenmiller:
That's a bit of the problem... not sure where to start with the changes needed? Any idea who might be able to provide guidance?

omehegan:
Maybe @mreichel or @colourmeamused.

mreichel:
@acohen716 @bbodenmiller How is your enterprise Jenkins setup? Do you use the Folders plugin in combination with the RBAC plugin?

bbodenmiller:
Yep, we're using the [CloudBees Folders Plugin](https://www.cloudbees.com/products/cloudbees-jenkins-platform/team-edition/features/folders-plugin) and [CloudBees Role-based Access Control Plugin](https://www.cloudbees.com/products/cloudbees-jenkins-platform/team-edition/features/role-based-access-control-plugin). Please let me know what other details about our setup would be useful to know about.

mreichel:
[...]

PR #916, opened 2019-04-10
Feature #631 : Allow for Jenkins project/folder level GitLab Token API credentials to be used for GitLab connection

mat1e:
I create this pull request to answer to the issue https://github.com/jenkinsci/gitlab-plugin/issues/631

No change have been made on GitLab Global configuration, but now you don't necessary need to add a credential in the global context :

Capture_global_credential

You can add one directly on the Job :

Capture_credential_folder

**Like this, gitlab-plugin is able to use a folder scope credential.**

If a credential is set on both configuration (Global level and Job level), the one of the job will be taken.
If no credential have been set on the job, the global one will be taken.

If you have any ideas or suggestions, you can add a comment to this PR :)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,7 +17,7 @@
   </properties>
 
   <artifactId>gitlab-plugin</artifactId>
-  <version>1.5.12-SNAPSHOT</version>
+  <version>1.5.13-SNAPSHOT</version>
   <name>GitLab Plugin</name>
   <url>https://wiki.jenkins-ci.org/display/JENKINS/GitLab+Plugin</url>
   <packaging>hpi</packaging>
--- a/src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnection.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnection.java
@@ -111,17 +111,17 @@ public class GitLabConnection {
         return readTimeout;
     }
 
-    public GitLabClient getClient() {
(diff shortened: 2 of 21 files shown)


ipleten:
Looks like that MR has broken compatibility with existing pipelines:
I build plugin fro master and now getting error:

> "gitLabConnection" should have 3 arguments but has 1 arguments instead. @ line 58, column 7.
>          gitLabConnection('gitlab.example.com')

How it supposed to be used with Global config:

` gitLabConnection('gitlab.example.com',null,null)`
?
If so it will break all existing pipelines.

mat1e:
> 
> 
> Looks like that MR has broken compatibility with existing pipelines:
> I build plugin fro master and now getting error:
> 
> > "gitLabConnection" should have 3 arguments but has 1 arguments instead. @ line 58, column 7.
> > gitLabConnection('gitlab.example.com')
> 
> How it supposed to be used with Global config:
> 
> ` gitLabConnection('gitlab.example.com',null,null)`
> ?
> If so it will break all existing pipelines

Thank you for your report ! 
Can you create an issue for this and post the number here? I make the correction so that the old method works again. 

For now, the missed parameters are "useAlternativeCredential" and "jobCredentialId". To make your pipeline work must do :
`gitLabConnection(gitLabConnection: 'gitlab.example.com', useAlternativeCredential: false, jobCredentialId: null)`
Like this it will use the Global config.

PR #923, opened 2019-05-01
Support user regex filter for note trigger

the-sea:
Support user regex filter when trigger type is note.If the username who wrote comment like "merge please" not match the user regex configured in job configuration, the note will not trigger build.

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/trigger/handler/note/NoteHookTriggerHandlerImpl.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/trigger/handler/note/NoteHookTriggerHandlerImpl.java
@@ -1,7 +1,10 @@
 package com.dabsquared.gitlabjenkins.trigger.handler.note;
 
 import com.dabsquared.gitlabjenkins.cause.CauseData;
+import com.dabsquared.gitlabjenkins.connection.GitLabConnectionProperty;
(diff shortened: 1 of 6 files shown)


atrifex:
Why hasn't this been reviewed and checked in? @omehegan can you please take a look.

src/main/java/com/dabsquared/gitlabjenkins/trigger/handler/note/NoteHookTriggerHandlerImpl.java
```diff
@@ -67,8 +73,8 @@ protected CauseData retrieveCauseData(NoteHook hook) {
                 .withTargetProjectId(hook.getMergeRequest().getTargetProjectId())
                 .withBranch(hook.getMergeRequest().getSourceBranch())
                 .withSourceBranch(hook.getMergeRequest().getSourceBranch())
-                .withUserName(hook.getMergeRequest().getLastCommit().getAuthor().getName())
-                .withUserEmail(hook.getMergeRequest().getLastCommit().getAuthor().getEmail())
+                .withUserName(hook.getUser().getUsername())
```
atrifex, line 76: This is not backward compatible. People that are relying on this being the user name and email of committer will see bugs. 

Might be useful to create new field that deals with the hook user.

PR #940, opened 2019-06-20
WIP: Retrieve merge request status for any job

derrickgw:
This merge request adds a pipeline function `gitlabMergeRequestStatus(projectId, branchName)` to retrieve merge request information about any given project/branch.

This would be useful in cases where the pipeline needs to behave differently if there is a merge request open, e.g. running long resource intensive tests only for merge requests, or in pipelines that are triggered by other jobs.  This is a partial solution to the fact that Multi-branch pipelines don't get run for merge requests.  These values are retrieved directly since they can't be propagated to the multi-branch build.

Ideally this could be tested by having a mock gitlab instance provide canned JSON responses, but I don't know how to accomplish that within the existing framework.

This is WIP to:
* Gauge interest
* Get code review
* Get advice on how to test


markjacksonfishing:
@derrickgw did you run a `mvn clean install`?

PR #774, opened 2018-06-05
fix commit status ref on tag push event

ScienJus:
fix #677 .

When a tag push event is received, `updateGitlabCommitStatus` will use the tag name instead of the branch name to ensure consistency with gitlab ci.

Issue #677, opened 2017-11-29
Pipeline ref name incorrect for tag push events


gempain:
### Context
- **Gitlab plugin version**: 1.5.2
- **Gitlab version**: 10.2.2
- **Jenkins version**: 2.73.3
- **Job type**: Freestyle

### Problem description

When a build is triggered by a tag push event, the CI job branch name that appears in Gitlab CI is wrong:
![pipeline](https://user-images.githubusercontent.com/13135149/33368255-76199968-d4f1-11e7-8acc-a1f24fa7821c.png)
I would expect `refs/tags/<tag-name>` to be `<tag-name>`.

This has two consequences:
- it conflicts with Gitlab CI native behavior: for a tagged commit, CI jobs use the tag name as job branch name:
![pipeline_gitlab_ci](https://user-images.githubusercontent.com/13135149/33368463-0cf05746-d4f2-11e7-8743-e7fdb99eb5e7.png)
- when you click the title, it should take you to `<gitlab-url>/<repo-path>/commits/<tag-name>`, but it actually takes you to `<gitlab-url>/<repo-path>/commits/refs/tags/<tag-name>`, which returns a 404.

Edit: here is an even better example:
<img width="993" alt="screen shot 2017-12-01 at 00 58 50" src="https://user-images.githubusercontent.com/13135149/33461466-e087907e-d632-11e7-8b7a-2f0014fa0e05.png">

omehegan:
I suspect this is an issue with how we return data to GitLab for builds triggered by tags. Needs investigation.

omehegan:
Released in version 1.5.7.

rililinx:
Hi. I faced with it in:
 gitlab plugin version: 1.5.10
Gitlab version: 11.2.3
Jenkins version: 2.138.2
Job type: Freestyle

PR #944, opened 2019-07-11
Fix tag ref when setting build status.

jukov:
I have same issue described in #677.

After review pull #774 i figured out not only CommitStatusUpdater update build status, and remove branch ref in other places.

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/trigger/handler/AbstractWebHookTriggerHandler.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/trigger/handler/AbstractWebHookTriggerHandler.java
@@ -70,9 +70,10 @@ public abstract class AbstractWebHookTriggerHandler<H extends WebHook> implement
                 if (client == null) {
                     LOGGER.log(Level.SEVERE, "No GitLab connection configured");
                 } else {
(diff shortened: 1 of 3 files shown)


thomas955:
We had the same issue with Jenkins CI -> and Tag pushing from gitlab.
Like discussed here and in https://github.com/jenkinsci/gitlab-plugin/issues/677 the behaviour is 100% the same. 

I compiled the gitlab plugin with the changes from jukov and for me its working so far. it would be nice if this could be merged as soon as possible into the official gitlab-jenkins plugin

Thanks @jukov

p.s. there is a stackoverflow post with the same problem
https://stackoverflow.com/questions/52933087/double-pipelines-on-gitlab-tag-webhook

rlfnb:
@omehegan Would be great to see it merged, still running into the same problem being described and already fixed in the PR sent by @jukov

markjacksonfishing:
@jukov Is there an associated Github issue with this PR? What about a Jenkins jira? If so, can you link them? If not, can you open them and note this PR?
Thank you for that.

thomas955:
@omehegan Is it possible to merge this into the master soon? Would be nice.
 Thank you in advance.

thomas955:
Just created a ticket in jira: see https://issues.jenkins-ci.org/browse/JENKINS-60233

PR #947, opened 2019-07-15
noteRegex help: add tips about (?m) and (?s) being useful

cben:
Followup to #628.

I'm coming here from a project where we configured noteRegex `^(\[test\]|\/retest)$` and were surprised it only triggers when completely alone in a comment, doesn't trigger for MR comments like:
```markdown
git fetch timed out, retrying.

/retest
```

I see noteRegex is fed to standard Java `.matches()`:
```
    final Pattern pattern = Pattern.compile(this.noteRegex);
    return pattern.matcher(note).matches();
```
There is presently no out-of-band way (and I don't think there should be) to specify options like `Pattern.MULTILINE` or `Pattern.DOTALL`.  
I believe matching multi-line comments is a common use case, so mentioning default `^`, `$`, `.` behavior and the way to sneak modifiers into the string can be helpful?

E.g. in our use case, turns out we want `(?sm).*^(\[test\]|\/retest)$.*`.

### DIFF
--- a/src/main/webapp/help/help-noteRegex.html
+++ b/src/main/webapp/help/help-noteRegex.html
@@ -3,6 +3,6 @@
         <p>When filled, a comment in the merge request completely matching this regular expression will trigger a build.
            Note this uses default <a href="https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html">Java regex syntax</a>.
            Notably <code>^</code> and <code>$</code> only match at the start/end of the whole comment, and <code>.</code> doesn't match newlines.
-           You might want to <a href="https://www.logicbig.com/tutorials/core-java-tutorial/java-regular-expressions/regex-embedded-flags.html">embed syntax modifiers</a> like <code>(?m)</code> <code>(?s)</code> to change these.</p>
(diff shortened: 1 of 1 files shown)


markjacksonfishing (changes requested):
Hi @cben and thank you for contributing.
So question of questions. 1) The 1st link references `Java SE 9 & JDK 9` which we are not using. That is not a huge deal because there is not much change there from java 8 (what we are currently using) to java 9 but I prefer to stay uniformed here.
2) Your 2nd link (https://www.logicbig.com/tutorials/core-java-tutorial/java-regular-expressions/regex-embedded-flags.html) is rather outdated (2016 creation date with 2017 update). While the page is overall correct I want to refrain from linking to a page I do not have approval to do so and especially when it has words like `Stew Pasta Twinkies` (ok, I laughed a little).
I love your tenacity and willingness to contribute and I want to accept this PR but lets do the following:
* Reference the correct java link
* Write a gist with your examples and link to that. This was I can give you credit for that work.

Thanks kindly!

cben:
Thanks for the review.
(1) done  
(2) I'm not really a Java programmer, and don't feel confident to write an "authoritative" post on it, much less maintain one over time.  => Linked to official Java tutorial instead.  
It's has less examples but only uses serious enterprise-safe words like `foo` :wink:

markjacksonfishing (approved):
/lgtm and awesome job @cben. I will be getting this in to the next release

Issue #891, opened 2019-02-03
Close merge request build are not triggered in version 1.5.10 and 1.5.11


haminhcong:
### Context

- Gitlab plugin version: 1.5.11 and 1.5.10
- Gitlab version: 10.8.6-ce
- Jenkins version: 2.150.2
- Job type: Pipeline

I have a CI system with jenkins and Gitlab. Three months ago, my boss upgade gitlab plugin from 1.5.9 to 1.5.10. A day later, i close a merge request and see that close merge request  event are not trigger any build, although I still see merge request hook still received and handled by Gitlab plugin.

Then i go to my PC and setup a demo environment with gitlab v10 and v11, with Jenkins Gitlab plugin version is 1.5.9, and merge request build work again. Therefore, i think that we has some problem with Jenkins Gitlab plugin version 1.5.10 and 1.5.11.

Can you help me resolve this problem ? Thank you very much.

Below is logs in Jenkins log from my test when i click "close merge request" button in gitlab.

#### With v1.5.10 and v1.5.11

```log
Feb 03, 2019 10:23:32 AM INFO com.dabsquared.gitlabjenkins.webhook.GitLabWebHook getDynamic

WebHook called with url: /project/spring_test_close_mr

Feb 03, 2019 10:23:32 AM FINE com.dabsquared.gitlabjenkins.webhook.build.MergeRequestBuildAction

MergeRequest: {
  "object_kind" : "merge_request",
  "event_type" : "merge_request",
  "user" : {
    "name" : "cicd_bot",
    "username" : "cicd_bot",
    "avatar_url" : "https://www.gravatar.com/avatar/4544bb20d88b39400e2b1a8262b9d875?s=80&d=identicon"
  },
  "project" : {
    "id" : 1,
    "name" : "spring_test",
    "description" : "",
    "web_url" : "http://gitlab:8180/cicd_bot/spring_test",
    "avatar_url" : null,
    "git_ssh_url" : "ssh://git@gitlab:8122/cicd_bot/spring_test.git",
    "git_http_url" : "http://gitlab:8180/cicd_bot/spring_test.git",
    "namespace" : "cicd_bot",
    "visibility_level" : 20,
    "path_with_namespace" : "cicd_bot/spring_test",
    "default_branch" : "master",
    "ci_config_path" : null,
    "homepage" : "http://gitlab:8180/cicd_bot/spring_test",
[...]

PR #951, opened 2019-07-18
[Fix] Accepted merge request event can not trigger Jenkins job

maxshine:
## Problem Description

While we are using gitlab-plugin as tiggers for Jenkins builds by Merge Request events from gitlab, we hit an issue that the builds can not be fired correctly by `merge` action of any Merge Request.  As I dig the root cause, I think I have some findings and make this change for this issue.

## Scenario and Issue Solution

The CI process of our team based on Jenkins + Gitlab is like this :
- When there is a MR opened, a build should be triggered
- When there is an update of opened MR, a build should be triggered
- When the MR is accepted (merged), a build should be triggered.

So, the cause of the issue we have, it the 3rd scenario of accepted MR. As `gitlab-plugin` has an logic to check if the build should be skipped because it is already done in [this codes](https://github.com/jenkinsci/gitlab-plugin/blob/cb04dad9cb32dd53c2a54083473c9c74f443810b/src/main/java/com/dabsquared/gitlabjenkins/trigger/handler/merge/MergeRequestHookTriggerHandlerImpl.java#L186), the acceptance event of MR can never trigger the builds due to it skipped.

The fix to this is simple, as `Action.merge` enumeration element is added in `skipBuiltYetCheckActions` set. 

Please kindly review the change and accept it if this makes sense.

Thanks

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/trigger/handler/merge/MergeRequestHookTriggerHandlerImpl.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/trigger/handler/merge/MergeRequestHookTriggerHandlerImpl.java
@@ -41,7 +41,7 @@ class MergeRequestHookTriggerHandlerImpl extends AbstractWebHookTriggerHandler<M
 
     private final boolean skipWorkInProgressMergeRequest;
     private final Predicate<MergeRequestObjectAttributes> triggerConfig;
-    private final EnumSet<Action> skipBuiltYetCheckActions = EnumSet.of(Action.open, Action.approved);
+    private final EnumSet<Action> skipBuiltYetCheckActions = EnumSet.of(Action.open, Action.approved, Action.merge);
     private final EnumSet<Action> skipAllowedStateForActions = EnumSet.of(Action.approved);
     private final boolean cancelPendingBuildsOnUpdate;
 


markjacksonfishing:
@maxshine Is there an associated Github issue with this PR? What about a Jenkins jira? If so, can you link them? If not, can you open them and note this PR?
Thank you for that.

TonyNoble:
Issues 891 and 958 partially address this issue.

https://github.com/jenkinsci/gitlab-plugin/issues/891
https://github.com/jenkinsci/gitlab-plugin/issues/958

It would be preferable if the ability to re-build previously-build commits was user-selectable, but at the very least, a merged MR should not be ignored purely because it had been built when the MR was opened in the first place.

maxshine:
> Issues 891 and 958 partially address this issue.
> 
> #891
> #958
> 
> It would be preferable if the ability to re-build previously-build commits was user-selectable, but at the very least, a merged MR should not be ignored purely because it had been built when the MR was opened in the first place.

Thanks @TonyNoble for adding the related issues. The two are similar to what I hit before.

maxshine:
Also, I think the gitlab-plugin gives the option for user to select which events could trigger the builds as :
Screen Shot 2019-10-12 at 11 05 38 AM

So my fix is just to ensure the selected option would work as expected.

interone-ms:
See PR #1001, I have made the check in isLastCommitNotYetBuild overridable on a job level.

zottinor:
I'm with the exactly issue on GitLab where Merge Accept button doesn't trigger jenkins job. This feature only works if a new branch was created from GitLab UI from Repository > Branches and used the green bottom New Branch. Starting a new branch from there when you click on Merge buttom at Merge Request open it trigger a new job on Jenkins.

Issue #926, opened 2019-05-14
Builds are triggered for non-code changes


jkugler:
## Issue
Builds get triggered on *any* update to the merge request, not just code pushes and comments.

### Context
- **Gitlab plugin version**: 1.5.10
- **Gitlab version**: 11.4.0
- **Jenkins version**: 2.175
- **Job type**: Matrix

### Logs & Traces
I will provide logs if needed.

### Problem description
In a prior version of the plugin, MR builds would only be triggered when the MR was created, when code was updated, or when the "magic comment" was made on the MR. As of a somewhat recent update (1.5.10, I'm not sure now), *any* change to the MR will trigger a build: change the assignee, change the title, change the description, etc.

_Describe your problem in a meaningful way_:
Making changes to the MR that should not trigger a build still triggers a build.

sladkoff:
Also happening with:

- **Gitlab plugin version**: 1.5.12
- **Gitlab version**: 11.10.4
- **Jenkins version**: 2.164

Minipada:
Any news on this? @jkugler

ababushk:
I think the problem is that GitLab sends an MR event on every change in a MR (description, assignee, labels, etc.) and plugin doesn't distinguish them. Looks like the following fields in JSON payload allow to check if actual commits were pushed:
```
    "action": "update",
    "oldrev": "0d6ea67a317dd6155425edc8b631f099dc0ec506"
```
So if both of these fields are present then there are code changes in MR.
I need this feature and I can develop it, but I'll certainly require some help because I'm new in Java and Jenkins plugins development

Issue #871, opened 2018-12-17
Labeling an open MR triggers a build


1100101:
## Issue
When making minor changes to an open merge request that should not affect the result of the Jenkins run, a rebuild is triggered still.
This includes adding/removing a label, setting/removing a milestone, as well as editing the description (or title) of the MR.

**Note:** This issue is closely related to issue #696, which was closed as a duplicate to #633, which did not fix the issues described here (or in #696).

### Context
- **Gitlab plugin version**: 1.5.5
- **Gitlab version**: 10.4.6 EE
- **Jenkins version**: 2.138.2
- **Job type**: Freestyle

### Logs & Traces

Here's a log extract from Jenkins.
The relevant part is the `changes` section. In this case, I only changed the Milestone assignment for the MR:
```
WebHook called with url: /project/navi-builder

Dec 17, 2018 8:37:36 AM FINE com.dabsquared.gitlabjenkins.webhook.build.MergeRequestBuildAction

MergeRequest: {
  "object_kind" : "merge_request",
  "user" : {
    "name" : "Frank Aurich",
    "username" : "f.aurich",
    "avatar_url" : ***FILTERED***
  },
  "project" : {
    "id" : 1242,
    "name" : ***FILTERED***,
    "description" : "",
    "web_url" : ***FILTERED***,
    "avatar_url" : null,
    "git_ssh_url" : ***FILTERED***,
    "git_http_url" : ***FILTERED***,
    "namespace" : ***FILTERED***,
    "visibility_level" : 10,
    "path_with_namespace" : ***FILTERED***,
    "default_branch" : "mw_develop",
    "ci_config_path" : null,
    "homepage" : ***FILTERED***,
    "url" : ***FILTERED***,
    "ssh_url" : ***FILTERED***,
    "http_url" : ***FILTERED***
  },
  "object_attributes" : {
    "assignee_id" : 316,
    "author_id" : 316,
    "created_at" : "2018-12-14 17:18:58 +0100",
    "description" : ***FILTERED***,
    "head_pipeline_id" : 61912,
    "id" : 26510,
    "iid" : 63,
    "last_edited_at" : "2018-12-14 17:19:30 +0100",
    "last_edited_by_id" : 316,
    "merge_commit_sha" : null,
    "merge_error" : null,
    "merge_params" : {
      "force_remove_source_branch" : "1"
    },
    "merge_status" : "can_be_merged",
    "merge_user_id" : null,
[...]

PR #903, opened 2019-03-06
Add possibility to force build with added MR label(s)

anton-johansson:
With this functionality (together with #901), I can perform logic using GitLab MR labels. For example, I can have a label titled "Auto-deploy", which I could use to automatically deploy my branch to my Kubernetes cluster.

This PR makes sure I can add the label "Auto-deploy" without having the GitLab Plugin block the webhook due to the fact that the Git SHA has already been built once.

PR #901 makes sure I can actually access the labels in the Jenkins pipeline, using environment variables.

PR #998, opened 2019-10-22
Trigger build only when new commits were pushed to MR

ababushk:
This PR resolves the problem described in #871 and #926 
An option in trigger configuration was added - when set build will be started on MR update only if it has new commits pushed to source branch. However, in current implementation build won't start when work in progress merge request become non-WIP MR.
I can use some help to write tests and fix the issue with WIP merge requests.

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/GitLabPushTrigger.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/GitLabPushTrigger.java
@@ -81,6 +81,7 @@ public class GitLabPushTrigger extends Trigger<Job<?, ?>> implements MergeReques
     private boolean triggerOnPush = true;
     private boolean triggerToBranchDeleteRequest = false;
     private boolean triggerOnMergeRequest = true;
+    private boolean triggerOnlyIfNewCommitsPushed = false;
     private boolean triggerOnPipelineEvent = false;
     private boolean triggerOnAcceptedMergeRequest = false;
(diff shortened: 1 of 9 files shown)


ababushk:
Hi @anton-johansson! Thank you very much for changes introduced in #903! They were exactly what I needed to finish this PR and I had no idea how to implement them by myself. If you have some spare time I humbly ask you to review this PR because I think it can interfere with your changes (I've tried to resolve conflicts as accurate as possible)

ababushk:
@anton-johansson it did interfere with your changes, so I slightly changed a condition and now MR will be built anyway if force build labels were set

slyalin:
Can we assign someone to review and merge changes? It is a very desired fix.

ababushk:
@markyjackson-taulia can you please take a look?

ababushk:
Thank you @markyjackson-taulia !

sandeepnagori87:
@ababushk was this issue "However, in current implementation build won't start when work in progress merge request become non-WIP MR." fixed as well?

I see that when we move MR from "WIP" to Ready, pipeline does not start

ababushk:
@sandeepnagori87 It's been a while, so I don't remember details, but according to https://github.com/jenkinsci/gitlab-plugin/blob/76afae8818251c9072fc8cb8a4c868902148c06d/src/main/java/com/dabsquared/gitlabjenkins/trigger/handler/merge/MergeRequestHookTriggerHandlerImpl.java#L260 it was fixed and should work.

Issue #950, opened 2019-07-18
GitLab plugin checks only global roles, project role is not verified


frluo:
#issue  
Per plugin instruction document, I try to use userid+token for Jenkins job authentication:
https://github.com/jenkinsci/gitlab-plugin#gitlab-to-jenkins-authentication, and use Configuring global authentication. 

First, I grant the user with job/Build permission to certain jobs, e.g. grant permission to a folder which has several jobs.
Second, I grant the user permission in global level, i.e. all jobs that has been assigned with job/Build permission.

After setting up above webhook in GitLab, and trigger the Job via GitLab, in the first scenario, the authentication is failed with code 403. and the second scenario is successful.

It seems GitLab plugin do not verify the folder level job/Build permission, instead, it checks only towards global job/Build permission.

### Context
- **Gitlab plugin version**: 1.5.12
- **Gitlab version**: 11.10.4
- **Jenkins version**: 2.176.1
- **Job type**: Pipeline

### Logs & Traces

<html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=utf-8"/>
<title>Error 403 xyz_user is missing the Job/Build permission</title>
</head>
<body><h2>HTTP ERROR 403</h2>
<p>Problem accessing /project/CTO/DevOps/sandbox/demo/test-global-role. Reason:
<pre>    xyz_user is missing the Job/Build permission</pre></p><hr><a href="http://eclipse.org/jetty">Powered by Jetty:// 9.4.z-SNAPSHOT</a><hr/>

</body>
</html>

in Jenkins log:
Jul 18, 2019 3:36:30 AM FINEST com.dabsquared.gitlabjenkins.webhook.build.BuildWebHookAction
Unauthorized (Did you forget to add API Token to the web hook ?)

### Problem description

_Describe your problem in a meaningful way_:
- what were you doing (simple push, merge request, MR with fork, ...)
Configure a Jenkins job webhook in gitlab in use global authentication way, i.e. with userid+API token+Jenkins Job URL. And trigger a push event from GitLab.
This userid has been assigned with Job/Build permission for certain folders in Jenkins.

- what was expected
The webhook is able to trigger Jenkins job.

- what occurred finally
The push event from GitLab is rejected with code 403, user is missing Job/Build permission.
When user assigned with global Job/Build permission, the Jenkins job can be triggered successfully.

Fearxpc:
+1
[...]

PR #1002, opened 2019-10-30
Check project item permission and not only global permission

Fearxpc:
The plugin checked only the global permissions.
But often people configure only folder or project permissions to have a more specific access control.
These were ignored.

This works also if a Jenkins instance uses only the global permissions because the project item inherits the permissions.

This commit fixes issue #950

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/webhook/build/BuildWebHookAction.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/webhook/build/BuildWebHookAction.java
@@ -47,7 +47,7 @@ abstract class BuildWebHookAction implements WebHookAction {
             GitLabPushTrigger trigger = GitLabPushTrigger.getFromJob((Job<?, ?>) project);
             if (trigger != null) {
                 if (StringUtils.isEmpty(trigger.getSecretToken())) {
-                    checkPermission(Item.BUILD);
+                    checkPermission(Item.BUILD, project);
                 } else if (!StringUtils.equals(trigger.getSecretToken(), secretToken)) {
                     throw HttpResponses.errorWithoutStack(401, "Invalid token");
                 }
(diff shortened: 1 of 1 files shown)


bbreijer:
Nice one!
I'am not a 100% sure it will lead to the desired result, did you test it on a local instance?

The reason why i'am having doubts is that the given authentication is used to verify the project permission, but this authentication is impersonated using `ACL.SYSTEM` in for instance https://github.com/jenkinsci/gitlab-plugin/blob/e66ccc0fb76d335d5d28945359023db938a6cc24/src/main/java/com/dabsquared/gitlabjenkins/webhook/build/PushBuildAction.java#L82-L90

Do you agree?

I'am thinking of using the committers username at that point to check the permissions. @markyjackson-taulia how do you feel about that?

Fearxpc:
Thank you for your response.

> I'am not a 100% sure it will lead to the desired result, did you test it on a local instance?

Yes i made a couple of test which were sucessfull.

Created a folder "test" with a project "gitlab_test" inside.
Created dummy user (non admin).

1. Jenkins without any permissions
-> Gitlab triggered the job ✔️ 
2. Jenkins with activated matrix based security (set no permissions)
-> Job was **not** triggered ✔️ 
3. Added Build permission to authenticated users (global permission)
-> Job was triggered ✔️ 
4. Added Build permission to dummy user only (global permission)
-> Job was triggered ✔️ 
5. Added Build permission to dummy user on the folder only
-> Job was triggered ✔️ 
6. Added Build permission to dummy user on the project only
-> Job was triggered ✔️ 


> The reason why i'am having doubts is that the given authentication is used to verify the project permission, but this authentication is impersonated using ACL.SYSTEM in for instance

At the moment I see what you mean but I can't give a good answer to it yet. :) Need to dig a little bit more into the code for it.

Fearxpc:
So I made some more tests today and checked the code.

> The reason why i'am having doubts is that the given authentication is used to verify the project permission, but this authentication is impersonated using `ACL.SYSTEM` in for instance
> 
> https://github.com/jenkinsci/gitlab-plugin/blob/e66ccc0fb76d335d5d28945359023db938a6cc24/src/main/java/com/dabsquared/gitlabjenkins/webhook/build/PushBuildAction.java#L82-L90

The code impersonates the `ACL.SYSTEM` to call the `TriggerNotifier` method `run()` and the following code but the authentication itself is made with the project ACL object.

https://github.com/jenkinsci/gitlab-plugin/blob/929bb5008866d2cca966cc4e0672420716f350e1/src/main/java/com/dabsquared/gitlabjenkins/webhook/build/BuildWebHookAction.java#L60-L64

`project.getACL()` returns the ACL object of the job/project item and the `authentication` object contains the user which was used to create the webhook in Gitlab.

So yes in all tests I got the expected result :)

bbreijer:
Since we are only using the project specific configuration (https://github.com/jenkinsci/gitlab-plugin/blob/e66ccc0fb76d335d5d28945359023db938a6cc24/README.md#L121), i did not consider the configuration mentioned in  (https://github.com/jenkinsci/gitlab-plugin/blob/e66ccc0fb76d335d5d28945359023db938a6cc24/README.md#L114). For this configuration, your PR works perfectly!

I tested everything using the `Global configuration` method, and it makes perfect sense and there are no doubts left for me ;)

PR #1032, opened 2020-02-26
Set commit status to 'canceled' only when user aborts build from UI

### LINKED ISSUE #1031: Plugin sets status 'Canceled' instead of 'Failed' when failed job calls downstream builds with Pipeline's 'build' step
### Issue
Plugin marks commit as `canceled` instead of `failed` if triggered Pipeline job starts downstream builds with `build` step and `propagate: true`, and they fail

### Context
- **Gitlab plugin version**: 1.5.17 (actually it is built from master branch, commit sha 05806187a89b67332c4bf3db56c204052d38dac9)
- **Gitlab version**: 11.11.7
- **Jenkins version**: 2.190.1 from the plugin POM and 2.204.2
- **Job type**: scripted Pipeline
- **Pipeline: Build Step plugin**: 2.11

### Reproducer Jenkinsfile
```
gitlabCommitStatus(name: "pre_commit")
                    {
                        echo "Hello, world!"
                        sleep 10
                        build job: "thisjobwillfail", propagate: true
                    }
        
```

### Problem description
After https://github.com/jenkinsci/pipeline-build-step-plugin/pull/24/ was merged to to Pipeline: Build Step plugin (this was releases with version 2.10), GitLab plugin sends `canceled` status to a commit if downstream builds were started from a job triggered by it, and then failed.

ababushk:
Fixes #1031

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -190,7 +190,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-step-api</artifactId>
-      <version>2.20</version>
+      <version>2.22</version>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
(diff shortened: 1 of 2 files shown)


ababushk:
@markyjackson-taulia could you please take a look? Looks like some infrastructure issue involved.

ababushk:
@markyjackson-taulia we're using gitlab-plugin built from this PR successfully for over a month now, I think it may help other people too, could you please take a look?

markjacksonfishing (approved):

Issue #1010, opened 2019-12-12
gitlabCommitStatus pushing status to wrong branch


vkmellon:
### Context
- **Gitlab plugin version**: 1.5.13
- **Gitlab version**: 12.5.3-ee
- **Jenkins version**: 2.190.1` **Job type**: Multi branch Pipeline

### Problem description
When we are creating new branch (without pushing commits) - for example TestBranch from Develop it triggers build on TestBranch. But gitlabCommitStatus is reporting to Develop branch, leaving TestBranch without status.

Here is my test pipeline
```
properties([gitLabConnection('MyGitLabConnection')])
node('linux') {
  stage('Test') {
    gitlabCommitStatus('Test') {
      println('Test')
    }
  }
}
```

DieAkuteSense:
### Context
* **Gitlab plugin version**: 1.5.12
* **Gitlab version**: 13.0.5
* **Jenkins version**: 2.176.2
* **Job Type**: Multibranch PIpeline


We face a similar issue:

Using Merge Requests in GitLab with setting "Fast forward merge" (without merge request) we run into the same behaviour. When we do not squash the commits in the GitLab MR, the Commit Hashes in the source branch will be identical as they are in the target branch.
The Jenkinsfile notifies GitLab about the build progress using ``updateGitlabCommitStatus``. Due to the identical commit hashes the build progress will be mapped to the *source* Branch in GitLab, even if this branch is already deleted; not the *target* branch.
We can see this also in GitLab Merge Request View: The first pipeline (above the "Merge" button) runs before the merge - until now everything works fine. The 'second' pipeline (= the one of the target branch; below the "Merge" button) will only run if we use squash commits; otherwise the first pipeline runs again.

DieAkuteSense:
I think this is related to #704 (respectively #625 ) . The logic getting the source branch relies on getting a `GitLabWebHookCause`. In a multibranch setup, the build is only triggered by a `BranchIndexingCause` not containing any specific Git / GitLab data.
In an environment using another Jenkins setup this maybe works fine....

krisstern:
We have a PR #1288 purporting to fix this issue, but the contributor cannot test the proposed solution on a production server. I am wondering if anyone has run into this same issue in production and is willing to give it a test? Any help would be much appreciated.

PR #1068, opened 2020-06-17
Add Mock return and test correctly set ref in CommitStatusUpdaterTest

DieAkuteSense:
Investigating a possible cause of #1010 I've noticed a missing mock return in ``CommitStatusUpdaterTest``. Related to this, the tests verified that ``ref=null``, which does not align with the given mock values for ``causeData``, where the source branch is set to "feature". I've added the missing mock return statement and improved the tests to check for the correctly set ``ref="feature"``.

Unfortunately I have not found a cause for #1010 😞 

IntelliJ modified the imports a lot...so the diff looks very confusing.


MarkEWaite (approved):
I resolved the issues with the imports and adapted the pull request to spotless formatting.

Test change is ready to merge.

PR #1102, opened 2021-02-02
Fix dependencies and tests

zbynek:
Since the Jackson plugin imports Jackson 2 databind, it shouldn't be added directly.


zbynek:
#1103 fixes all the tests

PR #1103, opened 2021-02-05
Fix dependencies and mock to make unit test Success

bulanovk:
Fix dependencies and mock to make unit test Success

Changes were made:
1. Fix pom.xml to make pom.xml valid
2. Downgrade guava to 18.0 as it managed by Jenkins Core
3. Add mock in GitLabCommitStatusPublisherTest to avoid NPE
4. Bump jenkins-test-harness
5. Add timeout to avoid Temp Directory dispose error
6. Bump Jackson to Modern version with ordering bugfix 2.12.1
7. Update parent.pom to 4.16
8. Switch version management to DependenciesManagement using Jenkins BOM
9. Align all depends according to bom-2.222.x version 23
10. Minimal required version of Jenkins 2.224.4
11. Use README from repo as a plugin documentation on plugins.jenkins.io (Thanks to @MarkEWaite )

PR #1106, opened 2021-02-12
Fix master branch compile

MarkEWaite:
## Fix compile on master branch

- Apply pom.xml from develop to master to fix duplicate groupId
- Revert "Update pom.xml" to undo guava upgrade (guava is provided by Jenkins core, can't be upgraded)


MarkEWaite:
Closing in favor of #1103 because it does a much better job of addressing the problem.

PR #1108, opened 2021-02-17
Make field name match getter/setter

daniel-beck:
This fixes a Javadoc error by making field and getter/setter names match.

Unsure what the side effects are, might be better for now to ignore Javadoc failures.


omehegan:
It turned out that this was fixed with just https://github.com/jenkinsci/gitlab-plugin/commit/110f84ea84b56e5006e63e3b149a6d8a4fca960e (or so it appears...)

PR #1109, opened 2021-02-19
[JENKINS-64341] Remove the table, tr and td tags from the forms that should not be tables

ironcerocloudbees:
[JENKINS-64639]

I checked the master version running from a Jenkins 2.279 and I found everything is working as in Jenkins 2.263.4 it did. However, there is one jelly file where the table tags were been used to draw fields in the configuration forms. Those table tags are that I removed in the new Jenkins version using the variable divBasedFormLayout.

cc @timja, @fqueiruga, @dariver


timja (approved):
this plugin does have an issue but it was fixed awhile back on master but hasn't been released in a long time

fqueiruga (approved):

oleg-nenashev (approved):

oleg-nenashev:
@MaximeMazet Hi, are you the new plugin maintainer? It would be great to release this change so that the plugin is compliant with the new configuration layouts in Jenkins 2.266+

PR #1125, opened 2021-04-06
Update CHANGELOG

ljackiewicz:
I tried to include Pull Requests containing more significant changes for the end user. It seems that recent changes have been added in a similar way. Hope this will be helpful for users updating the plugin version on their Jenkins instance.

One of the PR was mistakenly listed under 1.5.14, when in fact it was added in 1.5.16.


ghost:
Thanks very much is merged ;)

PR #1146, opened 2021-07-09
prevent a NPE when calling 'ObjectId.fromString(scmRevisionHash)'

stmlange:
This MR avoid the following NPE:
```
18:13:18  [Pipeline] updateGitlabCommitStatus
18:13:18  Error when executing success post condition:
18:13:18  java.lang.NullPointerException
18:13:18  	at org.eclipse.jgit.lib.ObjectId.fromString(ObjectId.java:202)
18:13:18  	at com.dabsquared.gitlabjenkins.util.CommitStatusUpdater.retrieveGitlabProjectIds(CommitStatusUpdater.java:191)
18:13:18  	at com.dabsquared.gitlabjenkins.util.CommitStatusUpdater.updateCommitStatus(CommitStatusUpdater.java:61)
18:13:18  	at com.dabsquared.gitlabjenkins.util.CommitStatusUpdater.updateCommitStatus(CommitStatusUpdater.java:97)
18:13:18  	at com.dabsquared.gitlabjenkins.workflow.UpdateGitLabCommitStatusStep$UpdateGitLabCommitStatusStepExecution.run(UpdateGitLabCommitStatusStep.java:79)
18:13:18  	at com.dabsquared.gitlabjenkins.workflow.UpdateGitLabCommitStatusStep$UpdateGitLabCommitStatusStepExecution.run(UpdateGitLabCommitStatusStep.java:63)
18:13:18  	at org.jenkinsci.plugins.workflow.steps.AbstractSynchronousStepExecution.start(AbstractSynchronousStepExecution.java:42)
18:13:18  	at org.jenkinsci.plugins.workflow.cps.DSL.invokeStep(DSL.java:319)
18:13:18  	at org.jenkinsci.plugins.workflow.cps.DSL.invokeMethod(DSL.java:193)
18:13:18  	at org.jenkinsci.plugins.workflow.cps.CpsScript.invokeMethod(CpsScript.java:122)
18:13:18  	at sun.reflect.GeneratedMethodAccessor5043.invoke(Unknown Source)
18:13:18  	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
18:13:18  	at java.lang.reflect.Method.invoke(Method.java:498)
18:13:18  	at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:93)
18:13:18  	at groovy.lang.MetaMethod.doMethodInvoke(MetaMethod.java:325)
[...]

PR #1152, opened 2021-08-16
old buildata hash can be null - fix NPE

Fearxpc:
Like #1146 it looks like that in some cases the scm build data can also return a NPE which failed all of our builds.


brd:
@MaximeMazet Can this be merged too?

Fearxpc:
@basil can you please have a look at this and also #1146 
We are currently helped us by creating a custom build but we want to use offical releases again :) thx

basil (approved):

basil:
Released in [1.5.23](https://github.com/jenkinsci/gitlab-plugin/releases/tag/gitlab-plugin-1.5.23).

Issue #1200, opened 2021-12-06
Latest commit is not set in gitlabAfter param if multiple commits while build is running 


marcusphi:
### Version report

Jenkins and plugins versions report:

```
Jenkins: 2.303.1
OS: Linux - 3.10.0-1160.6.1.el7.x86_64
---
...
git:4.8.2
git-changelog:3.11
git-client:3.9.0
git-server:1.10
...
gitlab-api:1.0.6
gitlab-branch-source:1.5.9
gitlab-oauth:1.12
gitlab-plugin:1.5.24
...
scm-api:2.6.5
...
```

### Reproduction steps
- We have a Jenkins job that does not allow concurrent builds
- It is triggered by a gitlab webhook on push action
- While a build is running, push two times to the repo with the webhook.

### Results

Expected result:

If multiple pushes happens while a build is running, only one build is queued, but when it runs the `gitlabAfter` param is set to commit of **last** push. That is, the latter trigger should supersede the former.

Actual result:

`gitlabAfter` is set to commit of **first** push. 
No build at all is triggered for second push. That change is "pending" until a build is triggered next time.

I believe this is incorrect behaviour.

I also commented in https://github.com/jenkinsci/gitlab-plugin/issues/841 which sounds similar, but I wanted to have a new clean issue for this problem.

marcusphi:
Here is the Jenkins XML for the build when it's finished (somewhat shorted):

An interesting observation is that the `BuildData` action contains the last commit and the `changeSet` contains the penultimate, the first commit of the two. 

`gitlabAfter` in build is f8b182c...

Another interesting observation is that there are **two** GitLabWebHookCause actions. These correspond to the two pushes.

```
<freeStyleBuild _class="hudson.model.FreeStyleBuild">
	<action _class="hudson.model.CauseAction">
		<cause _class="com.dabsquared.gitlabjenkins.cause.GitLabWebHookCause">
			<shortDescription>Started by GitLab push by Marcus Philip</shortDescription>
		</cause>
		<cause _class="com.dabsquared.gitlabjenkins.cause.GitLabWebHookCause">
			<shortDescription>Started by GitLab push by Marcus Philip</shortDescription>
		</cause>
	</action>
	...
	<action _class="hudson.plugins.git.util.BuildData">
		<buildsByBranchName>
			<refsremotesoriginmaster _class="hudson.plugins.git.util.Build">
				<buildNumber>14</buildNumber>
				<marked>
[...]

PR #1230, opened 2022-01-21
#1200 combine commits so latest commit is used if multiple

marcusphi:
The problem and suggested solution is described in #1200.


marcusphi:
I think this is a bugfix for something that has been broken a long time. 
NB: No tests failed from this change. 
I will submit tests that fail with old code and passes with my change.

marcusphi:
This did not fix the problem ...

PR #1235, opened 2022-02-02
Upgrade Docker in GitLab

### LINKED ISSUE #1232: Docker GitLab version is old
### What feature do you want to see added?

We are currently using `sameersbn/gitlab:11.0.3` as the Docker GitLab image, which from 2018. GitLab now publishes an official Docker image which also ships with Postgres so we don't have to handle that ourselves. This will enable developers of this plugin to test against the latest GitLab version to ensure that everything works correctly.

### Upstream changes

_No response_

UnsolvedCypher:
Closes #1232 

Upgrade to the latest GitLab Omnibus Docker. This ensures that we can test the plugin against a modern version of GitLab, and also substantially simplifies the docker-compose setup.

### DIFF
--- a/src/docker/README.md
+++ b/src/docker/README.md
@@ -14,13 +14,13 @@ To start the containers, run `docker-compose up -d` from the `docker` folder. If
 
 ## Access GitLab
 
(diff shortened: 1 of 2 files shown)


basil (approved):
Nice! Would be great to also get the Docker-based Maven integration tests working again too. If you get it working locally I can hook it up on the GitHub/Jenkins side to run on all PR builds.

PR #1237, opened 2022-02-03
Remove @markyjackson-taulia from `FUNDING.yml`

basil:
@markyjackson-taulia is no longer active in maintaining this plugin.


markjacksonfishing (approved):
Yes, I agree. It was fun while it lasted but there are other people doing much more than I did and I am very thankful for them and the opportunity to have been a maintainer on this.

PR #1291, opened 2022-05-24
Remove unecessary dependencies

MarkEWaite:
## Remove unnecessary dependencies

Tests complete without a dependency on workflow-cps-global-lib.

PR https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/172 has deprecated workflow-cps-global-lib in
favor of pipeline-groovy-lib-plugin.  Remove the dependency because it is not required.

Adjust other dependencies to either remove them or make them `test` scoped.  Compiled and tested with Java 8 and Java 11.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -96,10 +96,6 @@
       <artifactId>workflow-api</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.jenkins-ci.plugins</groupId>
-      <artifactId>script-security</artifactId>
-    </dependency>
-    <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-step-api</artifactId>
     </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -89,10 +89,6 @@
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
-      <artifactId>structs</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>scm-api</artifactId>
     </dependency>
     <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -221,6 +221,13 @@
       <groupId>io.jenkins</groupId>
       <artifactId>configuration-as-code</artifactId>
       <scope>test</scope>
+      <exclusions>
+        <!-- Provided by mockserver when running in tests -->
+        <exclusion>
+          <groupId>org.apache.commons</groupId>
+          <artifactId>commons-lang3</artifactId>
+        </exclusion>
+      </exclusions>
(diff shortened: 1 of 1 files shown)


basil (approved):
Thanks for the PR!

Optional feedback for your consideration: Thinking more deeply about why this unused dependency was present in the first place, I suspect it was only added to deal with upper bounds errors (which is now irrelevant with the use of the plugin BOM). Running `mvn dependency:analyze` shows other unused dependencies as well, most likely originally added for the same (now irrelevant) reason. If desired, the scope of this PR could be expanded to remove these as well. The benefit of this would be to slim down the dependency tree of this plugin, reducing future maintenance burden. This is not a blocking comment, since this PR is an improvement over the status quo and is welcome as-is. The following additional unused dependencies were identified by `mvn dependency:analyze`:

- `junit`
- `script-security`
- `ssh-credentials`
- `structs`
- `workflow-api`
- `workflow-cps-global-lib`
- `workflow-durable-task-step`
- `workflow-scm-step`
- `workflow-support`

MarkEWaite:
I assume there a few alternatives to make the CI job pass.  

Options I see:

* Discard the changes that reduce the dependencies because they also seem to increase the spotbugs warnings.  Return to a single commit that removes a single dependency and rely on someone else to make any other improvements.  I believe the other improvements in this pull request are correct and complete, but the spotbugs failures due to 5 new spotbugs warnings are blocking the pull request
* Close the pull request and accept that I'll remove the gitlab plugin from my installation, accepting that it reduces exploratory testing in my environment
* Suppress spotbugs warnings execution on the plugin completely
* Spend hours suppressing newly introduced spotbugs warnings

Recommendations @basil?

basil:
I chose a variation of 3, since SpotBugs was basically ignored in this plugin anyway. In https://github.com/jenkinsci/gitlab-plugin/pull/1293 I changed `<spotbugs.failOnError>false</spotbugs.failOnError>` to `<spotbugs.threshold>High</spotbugs.threshold>` so that the "Quality Gate" won't interfere with unrelated PRs like yours. Thanks for the PR! Released in [1.5.33](https://github.com/jenkinsci/gitlab-plugin/releases/tag/gitlab-plugin-1.5.33).

pom.xml
```diff
@@ -87,25 +87,14 @@
...
       <artifactId>scm-api</artifactId>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-api</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.jenkins-ci.plugins</groupId>
-      <artifactId>script-security</artifactId>
+      <scope>test</scope>
```
jglick, line 97: Unlikely you need to specify transitive `test`-scoped dependencies like this either.

PR #1292, opened 2022-05-24
Remove JSR-305 dependency

MarkEWaite:
## Remove JSR-305 dependency

Jenkins 2.231 removed JSR-305 from the core so that it could be removed from use in Jenkins.

https://issues.jenkins.io/browse/JENKINS-55973 provides history of the change and the transition to use the spotbugs annotations instead of the JSR-305 annotations.  Since this plugin requires Jenkins 2.289.3 or newer, it can rely on the spotbugs annotations being available.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -153,11 +153,6 @@
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>display-url-api</artifactId>
     </dependency>
-    <dependency>
-      <groupId>com.google.code.findbugs</groupId>
-      <artifactId>jsr305</artifactId>
-      <version>3.0.2</version>
-    </dependency>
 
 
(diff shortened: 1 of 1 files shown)


basil (dismissed):
Thanks for the PR! I was surprised to see this still compiling because we still use `javax.annotation.Priority` in this plugin, but then I realized that `jcip-annotations-1.0-1.jar` is providing this class.

basil:
CI build does not compile.

basil:
I tried to tackle this a long time ago, but I never figured out the problem. Unless you have any ideas about to fix this, I suggest closing this PR in favor of letting this problem be resolved naturally by the requirement of Java 11.

MarkEWaite:
Agreed that closing this pull request is the best choice.  Thanks!

PR #1332, opened 2022-09-10
Optimize Gitpod config

jmMeessen:
Removes the multiple JDKs installations in Gitpod as Java17 is not a requirement anymore for VSC plugins.


gounthar (approved):
LGTM.
Thanks :pray:

PR #1346, opened 2022-10-19
Simplify jelly pages by removing divBasedFormLayout logic

MarkEWaite:
## Simplify jelly pages by removing divBasedFormLayout conditionals

Now that the plugin requires a Jenkins version newer than 2.277.1, there is no need for the divBasedFormLayout logic in the jelly forms.

### DIFF
--- a/src/main/resources/com/dabsquared/gitlabjenkins/GitLabPushTrigger/config.jelly
+++ b/src/main/resources/com/dabsquared/gitlabjenkins/GitLabPushTrigger/config.jelly
@@ -1,26 +1,9 @@
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core"
          xmlns:f="/lib/form"
-         xmlns:d="jelly:define"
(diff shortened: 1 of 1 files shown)


MarkEWaite:
@jmMeessen and @gounthar I will add this step to the ["Improve a plugin" tutorial](https://www.jenkins.io/doc/developer/tutorial-improve/) after I return from vacation.  The step is useful and reduces the maintenance burden of a plugin.  It requires a little more thought and analysis than some of the earliest steps and may appeal to those who want something a little more challenging.

Eventually I think that we'll want to add a step to modernize user interface components in plugins by using the guidance from the [Jenkins design library](https://plugins.jenkins.io/design-library/) as can be seen on [weekly.ci.jenkins.io](https://weekly.ci.jenkins.io/).

MarkEWaite:
Confirmed with interactive testing that the modified form displays as expected and remembers changes made in the form.  Ready to merge once CI passes.

PR #1353, opened 2022-11-08
Add spotbugs checks

MarkEWaite:
## Add spotbugs checks

- Remove unnecessary toString() call
- Use '%n' in format string, not '\n'
- Increase spotbugs checks, exclude specific warnings
- Suppress newly detected warnings


gounthar (approved):
Looks good to me. 👍 
Thanks. 😊

PR #1355, opened 2022-11-08
Add link to report an issue

MarkEWaite:
## Add link to report an issue

Increase chances that users will report issues to the correct location.


gounthar (approved):
Looks good to me. 👍 
Thanks. 😊

PR #1356, opened 2022-11-08
Require Jenkins 2.346.3 or newer

MarkEWaite:
## Require Jenkins 2.346.3 or newer

85% of users running one of the last two versions are using Jenkins 2.346.1 or newer.  Require Jenkins 2.346.3 or newer to reduce the maintenance load and narrow the testing.

- Remove jsr305 unnecessary dependency
- Get commons-io version from plugin bom
- Get asm version from plugin bom
- Get slf4j-api version from plugin bom
- Use latest bom release

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -313,7 +313,6 @@
       <dependency>
         <groupId>commons-io</groupId>
         <artifactId>commons-io</artifactId>
-        <version>2.11.0</version>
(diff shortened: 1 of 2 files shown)


gounthar (approved):
Looks good to me. 👍 
Thanks for the `JSR-305` too. 😊

PR #1360, opened 2022-11-15
Sort dependencies

basil:
This plugin has a huge number of dependencies. It's easier to read them when they're in some rough order. This PR sorts them by group ID, artifact ID, and scope. While I was here I also deleted any unnecessary dependencies from the `<dependencyManagement>` section and added an exclusion for JCIP annotations (which are no longer needed as of 2.326 or later because core now provides them).


MarkEWaite (approved):
Thanks!

PR #1363, opened 2022-11-22
Remove unneeded spotbugs exclusions

MarkEWaite:
## Remove unused spotbugs exclusions

Spotbugs has improved their detectors in the last two months.  These are no longer raised as warnings.


MarkEWaite:
@jmMeessen or @gounthar could you approve this change?  Consistent with the same change made in Jenkins core and in many plugins by Basil Crow.

gounthar (approved):
Looks good to me.
Thanks Mark!

PR #1371, opened 2022-12-02
Require Java 11

MarkEWaite:
## Require Java 11

Require Jenkins 2.361.4 or newer.

https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/ offers Jenkins 2.361.4 as one of the recommended versions.

Java 11 or newer is required for Jenkins 2.361.4 and for the most recent releases of the parent pom.


gounthar (approved):
Looks good to me.
Thanks.

PR #1394, opened 2023-02-01
Update Jenkins version to LTS 2.375.2

krisstern:
## Description
Update Jenkins version to LTS `2.375.2` in `pom.xml`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -8,22 +8,21 @@
     <relativePath />
   </parent>
   <artifactId>gitlab-plugin</artifactId>
-  <version>${revision}${changelist}</version>
+  <version>1.7.1</version>
   <name>GitLab Plugin</name>
   <url>https://github.com/jenkinsci/${project.artifactId}</url>
   <packaging>hpi</packaging>
 
   <properties>
-    <revision>1.7.2</revision>
+    <revision>1.7.1</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.361.4</jenkins.version>
+    <jenkins.version>2.375.2</jenkins.version>
     <spotbugs.effort>Max</spotbugs.effort>
     <spotbugs.threshold>Low</spotbugs.threshold>
     <gitHubRepo>jenkinsci/${project.artifactId}</gitHubRepo>
     <hpi.compatibleSinceVersion>1.4.0</hpi.compatibleSinceVersion>
     <mockserver.version>5.15.0</mockserver.version>
   </properties>
-
 
   <licenses>
     <license>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -8,23 +8,22 @@
     <relativePath />
   </parent>
   <artifactId>gitlab-plugin</artifactId>
-  <version>${revision}${changelist}</version>
+  <version>1.7.1</version>
```
MarkEWaite, line 11: This allows the publication of incrementals so that we can easily test a pull request from an environment that is managing plugin versions as code.
```suggestion
  <version>${revision}${changelist}</version>
```

pom.xml
```diff
@@ -8,23 +8,22 @@
...
   <name>GitLab Plugin</name>
   <url>https://github.com/jenkinsci/${project.artifactId}</url>
   <packaging>hpi</packaging>
 
   <properties>
-    <revision>1.7.2</revision>
+    <revision>1.7.1</revision>
```
MarkEWaite, line 17: Don't want to go backwards in time with version numbers
```suggestion
    <revision>1.7.2</revision>
```

pom.xml
```diff
@@ -8,23 +8,22 @@
...
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.361.4</jenkins.version>
+    <jenkins.version>2.375.2</jenkins.version>
```
MarkEWaite, line 19: The ["How to choose a base Jenkins version"](https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/) page recommends either 2.361.4 or 2.346.3 unless there are exceptional circumstances that recommend a newer version.  As far as I can tell, we should remain with 2.361.4 for now as our base Jenkins version.
```suggestion
    <jenkins.version>2.361.4</jenkins.version>
```

pom.xml
```diff
@@ -57,7 +56,7 @@
     <connection>scm:git:https://github.com/${gitHubRepo}.git</connection>
     <developerConnection>scm:git:git@github.com:${gitHubRepo}.git</developerConnection>
     <url>https://github.com/${gitHubRepo}</url>
-    <tag>${scmTag}</tag>
+    <tag>gitlab-plugin-1.7.1</tag>
```
MarkEWaite, line 59: This is used for the incremental build process.  We don't want to make this change.
```suggestion
    <tag>${scmTag}</tag>
```

MarkEWaite (changes requested):
@krisstern I think that we should close this pull request because we don't want to "lock out" users that are running the previous LTS.  Remaining with Jenkins 2.361.4 is also consistent with the recommendation in the [Jenkins documentation](https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/).

krisstern:
> @krisstern I think that we should close this pull request because we don't want to "lock out" users that are running the previous LTS. Remaining with Jenkins 2.361.4 is also consistent with the recommendation in the [Jenkins documentation](https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/).

Sure @MarkEWaite I actually wasn't sure about these changes so thanks for the documentation and advice!

pom.xml
krisstern: Sorry, there was a bug on GitHub before when I was trying to cut a new release, so unintentionally messed things up a bit.

Issue #1309, opened 2022-07-01
Restore detailed build cause description


daniel-beck:
### What feature do you want to see added?

https://github.com/jenkinsci/gitlab-plugin/commit/24e9a99d8151b5345109ef12cddc1ab323baa4ee implements a security fix: The previous, detailed build cause description rendered raw HTML containing some user-specified elements, causing [SECURITY-2316](https://www.jenkins.io/security/advisory/2022-06-30/#SECURITY-2316).

As this plugin is currently unmaintained (the [only person able to upload new releases](https://github.com/jenkins-infra/repository-permissions-updater/blob/c4b70c3d3f3a4db7da2aab2c685f5a1cf0748361/permissions/plugin-gitlab-plugin.yml#L11) [no longer has a GitHub account](https://github.com/jenkins-infra/repository-permissions-updater/pull/1983)/commit access) but very popular, the Jenkins security team decided to deliver a fix for this issue instead of [announcing it without a fix](https://www.jenkins.io/security/plugins/).

As we don't have the capacity or environment to test a "proper" fix, we opted to just remove all details from the build cause as a minimal, safe solution to address the vulnerability. This obviously isn't a great solution, and instead, build causes should provide all necessary details while properly escaping untrusted values from user input.

impact-maker:
@MarkEWaite I would like to work on this issue

krisstern:
Feel free to work on the issue @impact-maker

impact-maker:
Hi @krisstern  I have some issues building the code. Is there any slack or gitter channel for this project. Also I am not able to see any build instructions in the readme or the contributing guidelines.

krisstern:
Hi @impact-maker Sure, it is [here](https://gitter.im/jenkinsci/gitlab-plugin) on Gitter.

impact-maker:
@krisstern Since I have started with project from scratch. Some inputs would help to provide valuable contributions
I had the following queries in this issue.
1. How could I reproduce this error.
2. How could do the testing in case I make some code changes
Also the contributing guidelines are somewhat not clear. I can go ahead work on it as well after I am able to do it.

PR #1406, opened 2023-02-09
Reverted changes

impact-maker:
Description
Reverted back the changes in the build to fix issue #1309

### DIFF
--- a/src/main/resources/com/dabsquared/gitlabjenkins/cause/GitLabWebHookCause/description.jelly
+++ b/src/main/resources/com/dabsquared/gitlabjenkins/cause/GitLabWebHookCause/description.jelly
@@ -1,5 +1,5 @@
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core">
   <!-- SECURITY-2316: This used to show the HTML-formatted it.shortDescription, but that does not properly neutralize user-provided input -->
-  <span>Triggered by GitLab Webhook</span>
+  <span><j:out value="${it.shortDescription}" /></span>
 </j:jelly>


daniel-beck (changes requested):
Please do not introduce security vulnerabilities by reverting previous security fixes.

impact-maker:
> Please do not introduce security vulnerabilities by reverting previous security fixes.

Thanks for replying. Since I am a new contributer, could you help by guiding on how to reproduce issue #1309 and any insights to understand the issue better would be appreciated. Thank you

daniel-beck:
> how to reproduce issue

Unsure what you mean. If you're referring to [SECURITY-2316](https://www.jenkins.io/security/advisory/2022-06-30/#SECURITY-2316), put HTML in the payload of the webhook (which fields are included depends on the type of webhook) and see it rendered here.

krisstern:
Hi @daniel-beck I think there must have been some confusion. I looked and it seems #1309 has already been address previously in https://github.com/jenkinsci/gitlab-plugin/commit/24e9a99d8151b5345109ef12cddc1ab323baa4ee. Could you please confirm so that I can close both this PR and the corresponding issue?

krisstern:
@impact-maker I have just checked the issue was in fact a notice so the best we can do is to report it as completed instead of reverting to a former state to expose the previous vulnerabilities. But thanks for your code contribution!

daniel-beck:
> Hi @daniel-beck I think there must have been some confusion. I looked and it seems #1309 has already been address previously in [24e9a99](https://github.com/jenkinsci/gitlab-plugin/commit/24e9a99d8151b5345109ef12cddc1ab323baa4ee). Could you please confirm so that I can close both this PR and the corresponding issue?

I filed #1309 after I published https://github.com/jenkinsci/gitlab-plugin/commit/24e9a99d8151b5345109ef12cddc1ab323baa4ee, so no, it's not addressed. The issue is asking to restore the previous detailed build causes, but in a safe manner (i.e., not just undoing the security fix, as this PR did).

PR #1198, opened 2021-11-30
Fix access denied on test connection of a job for non admin users

mat1e:
Fix "Access Denied" error message when non admin user click on "Test Connection" button for the Gitlab connection on a Job.

PR #1439, opened 2023-03-29
Fix permission denied on test connection for non admin

mat1e:
See #1198 
Fix "Access Denied" error message when non admin user click on "Test Connection" button for the Gitlab connection on a Job.


krisstern (approved):
Thanks  @mat1e! Looks good.

daniel-beck:
This change makes no sense and should not be released. Explicitly checking for Overall/Read is extremely rarely needed (basically only in `UnprotectedRootAction`), since Jenkins takes care of this, and checking for this permission on an `Item` is never correct. I expect this will not work with non-trivial authorization strategies like `matrix-auth`, at least in some configurations.

Instead, check `item.hasPermission(Item.CONFIGURE)`. See https://www.jenkins.io/doc/developer/security/form-validation/#checking-permissions

krisstern:
Thanks @daniel-beck for the info, I have opened #1440 to correct this behavior.

mat1e:
Thanks for your review !

krisstern:
You are welcome @mat1e

PR #1449, opened 2023-04-11
Use upstream Spotless configuration

basil:
This PR applies the Spotless configuration from https://github.com/jenkinsci/plugin-pom/pull/733 to this repository and formats the entire repository by running `mvn spotless:apply`. See that PR for a description of the changes and their motivation. Since formatting is a subjective matter, the choice is up to the maintainers about whether or not they wish to opt in. I would not be upset if the maintainers choose not to accept this PR. If accepted, the next steps after merging this PR would be to add a `.git-blame-ignore-revs` file so that this commit does not show up in Git blame.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,6 +7,7 @@
     <version>4.59</version>
     <relativePath />
   </parent>
+
   <artifactId>gitlab-plugin</artifactId>
   <version>${revision}${changelist}</version>
   <packaging>hpi</packaging>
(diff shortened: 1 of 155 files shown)


basil:
The security scan issues are false positives, since those issues existed before this PR as well. It is just noticing them because the files are newly formatted and line numbers have changed.

krisstern (approved):
Look good

PR #1480, opened 2023-05-26
Fix Windows test by reverting to parent pom 4.63

MarkEWaite:
## Fix Windows tests by reverting to parent pom 4.63

Fix the test failures on Windows by reverting to previous parent pom.  Will need further investigation before we update this plugin to the most recent parent pom.

This reverts commit e940f618c03adb983e26add50ce6a8fc105a632e.


krisstern (approved):
LGTM

PR #1482, opened 2023-05-26
Clean build on project branch

basil:
Our CI job expects at least a single test to be run, so use `InjectedTest` for this purpose.


basil:
CC @krisstern

MarkEWaite (approved):

PR #1491, opened 2023-05-30
[GSoC Milestone 1] Migrate resource calls from RESTEasy to GitLab4J-API / Reference #1

harsh-ps-2003:
Migrating the resource calls from RESTEasy to GitLab4J-API without affecting the webhook implementation and Proxy settings not being functional.


src/main/java/com/dabsquared/gitlabjenkins/publisher/GitLabAcceptMergeRequestPublisher.java
```diff
@@ -56,13 +57,19 @@ public String getDisplayName() {
...
         try {
             if (build.getResult() == Result.SUCCESS) {
-                client.acceptMergeRequest(
-                        mergeRequest, "Merge Request accepted by jenkins build success", this.deleteSourceBranch);
+                gitLabApi
+                        .getMergeRequestApi()
+                        .acceptMergeRequest(
+                                mergeRequest,
+                                mergeRequest.getIid(),
+                                "Merge Request accepted by jenkins build success",
```
krisstern, line 65:
```suggestion
                                "Merge Request accepted by Jenkins build success",
```

basil:
PR description has very little context. Build is not clean either. I would be looking for both of those things before carefully reviewing this PR. Since it is likely going to be too much effort to deal with the tests in the short term, better to skip compiling and running them until `src/main` is more stable. I had already disabled execution of the tests on the project branch in #1482 but if they don't compile yet, let's skip compiling them too:

```xml
diff --git a/pom.xml b/pom.xml
index 776034e..6fc8db0 100644
--- a/pom.xml
+++ b/pom.xml
@@ -339,10 +339,9 @@
       -->
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-surefire-plugin</artifactId>
+        <artifactId>maven-compiler-plugin</artifactId>
         <configuration>
-          <test>InjectedTest</test>
-          <skipTests>true</skipTests>
+          <testExcludes>com/dabsquared/gitlabjenkins/**/*.java</testExcludes>
         </configuration>
       </plugin>
       <plugin>
```

The above should skip compilation and execution of our own tests while maintaining the dynmically generated `InjectedTest`, which we can also use as the one test we need for the `Jenkinsfile` build to pass (recall that the `Jenkinsfile` CI job requires at least one test to pass to mark the build as a pass).

With the above changes the build is almost clean, but there are some SpotBugs issues that need to be addressed. Let me know once you have a clean build in CI and a more thorough description of these changes and I will take a closer look.

README.md
```diff
@@ -47,7 +47,7 @@ This plugin allows GitLab to trigger builds in Jenkins when code is committed or
 This plugin is Open Source Software, developed on a volunteer basis by users of Jenkins and GitLab. It is not formally supported by either GitLab Inc. or CloudBees Inc.
 
 #### Supported GitLab versions
-GitLab performs a new major release about every six to nine months, and they are constantly fixing bugs and adding new features. As a result, we cannot support this plugin when used with GitLab versions older than N-2, where N is the current major release. At the time of this writing, the current stable release of GitLab is 11.1, so the oldest release supported by this plugin is 9.0.
+GitLab performs a new major release about every six to nine months, and they are constantly fixing bugs and adding new features. As a result, we cannot support this plugin when used with GitLab versions older than N-2, where N is the current major release. At the time of this writing, the current stable release of GitLab is 16.0, so the oldest release supported by this plugin is 14.0.
```
basil, line 50: The changes to the `README.md` are great! Very nice. I don't mind if they stay in this PR, but I also wouldn't mind if we filed a separate small PR with documentation updates (either to the main branch or the project branch, as appropriate). The smaller and more logically self-contained a PR is, the faster it will likely be merged.

src/main/java/com/dabsquared/gitlabjenkins/GitLabPushTrigger.java
```diff
@@ -187,53 +187,55 @@ public GitLabPushTrigger() {}
...
-                            10,
-                            10));
-
-            String defaultConnectionName = gitLabConfig.getConnections().get(0).getName();
-            for (AbstractProject<?, ?> project : Jenkins.getInstance().getAllItems(AbstractProject.class)) {
-                GitLabPushTrigger trigger = project.getTrigger(GitLabPushTrigger.class);
-                if (trigger != null) {
-                    if (trigger.addCiMessage) {
-                        project.getPublishersList().add(new GitLabCommitStatusPublisher("jenkins", false));
+        if (oldConfig != null) {
```
basil: I don't see how adding a null check here has anything to do with the rest of these changes. If this is desirable in and of itself please split it to a separate PR (either to the main branch or the project branch, as appropriate) including the reason why it is desirable and how it was tested.

src/main/java/com/dabsquared/gitlabjenkins/cause/CauseData.java
```diff
@@ -1,6 +1,11 @@
+/* Note for Reviewers :
+ * This Causedata will be broken according to the events received from Gitlab.
+ * gitlab4j uses Long data type instead of Integer
+ * as i cant create gitlabapi instance in causedata, getmergereqest has been moved to the classes which required in specifically.
+ */
+
```
basil: If this comment is describing the code, please put it next to the code it describes. If this comment is describing a change in the code (and not the code itself), please remove it from the Git changeset and instead leave it as a GitHub review comment (you can "review" your own PRs). If the former (that is, this is describing the code) and it is describing temporary code, then please mark it with `TODO` so that it is clearly obvious that the code being described is incomplete when comparing the diff between the main branch and the project branch.

src/main/java/com/dabsquared/gitlabjenkins/cause/CauseData.java
```diff
@@ -421,25 +426,6 @@ public String getMergeRequestAssignee() {
         return mergeRequestAssignee;
     }
 
-    @Exported
```
basil, line 424: `@ExportedBean` usually indicates that something is visible in the XML/JSON API, so you'll need to do some research with the old code to see where this exported data was being exposed in the XML/JSON API and ensure that the same data is exposed in the XML/JSON API after this PR.

src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnection.java
```diff
@@ -1,8 +1,11 @@
+/* Note for Reviewers :
+ * Proxy settings will be configured here, not implimented currently
+ * feature of filling clientbuilderid by user and the respective constructors has been removed
+ */
+
```
basil: Ditto about putting comments next to the code they describe, comments about code change in the PR description rather than the diff, and `TODO` next to comments about incomplete code.

src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnection.java
```diff
@@ -53,65 +55,32 @@ public class GitLabConnection extends AbstractDescribableImpl<GitLabConnection>
...
-                readTimeout);
-    }
+    private transient Map<String, GitLabApi> gitLabApiCache;
 
     @DataBoundConstructor
     public GitLabConnection(
             String name,
             String url,
             String apiTokenId,
-            String clientBuilderId,
```
basil: As described in https://github.com/jenkinsci/gitlab-plugin/pull/1491#discussion_r1210826532, we cannot break compatibility nor is it desirable to drop support for autodetection. This should be reverted.

src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnection.java
```diff
@@ -142,24 +107,24 @@ public int getReadTimeout() {
...
-            clientCache.put(
-                    clientId,
-                    clientBuilder.buildClient(url, token, ignoreCertificateErrors, connectionTimeout, readTimeout));
+        if (!gitLabApiCache.containsKey(gitLabApiId)) {
+            gitLabApiCache.put(gitLabApiId, createGitLabApi(url, token));
         }
-        return clientCache.get(clientId);
+        return gitLabApiCache.get(gitLabApiId);
+
+        // proxy configuration setup has to be done here
```
basil: If a comment describes incomplete code than please add `TODO`

src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnection.java
```diff
@@ -142,24 +107,24 @@ public int getReadTimeout() {
...
             // Add prefix to credential ID to avoid collision with "global"
-            clientId = "alternative-" + jobCredentialId;
+            gitLabApiId = "alternative-" + jobCredentialId;
             token = getApiToken(jobCredentialId, item);
         }
 
-        if (!clientCache.containsKey(clientId)) {
-            clientCache.put(
-                    clientId,
-                    clientBuilder.buildClient(url, token, ignoreCertificateErrors, connectionTimeout, readTimeout));
```
basil: What about `ignoreCertificateErrors` and `connectionTimeout`?

src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnection.java
```diff
@@ -53,65 +55,32 @@ public class GitLabConnection extends AbstractDescribableImpl<GitLabConnection>
     private transient String apiToken;
     // TODO make final when migration code gets removed
     private String apiTokenId;
-    private GitLabClientBuilder clientBuilder;
```
basil: This hunk should be reverted I think. The existing code is serializing to disk an XML object with an ID (either 'autodetect', 'v3', or 'v4') and an ordinal (integer version). We cannot change this existing format; see https://www.jenkins.io/doc/developer/persistence/backward-compatibility/ for more information.

Yes I understand that v3 support was removed from GitLab 11. At the same time, v5 support will be added in the future. And existing users may have the value set to 'autodetect', which tries each endpoint until it finds one that is working. And I think we want to continue to support this. For example, in the future, GitLab X will have support for v4 and v5, at which point this plugin would want to support that while retaining support for GitLab X-1 which only supports v4. My point is that the autodetect mode is valuable and should not be ripped out.

To keep support for the autodetect mode with GitLabApi, you'd want to enhance whatever method creates the `GitLabApi` object (in the current PR, `createGitLabApi`) to iterate through all API versions in `org.gitlab4j.api.GitLabApi.ApiVersion` and try each one until finding one that works. For example (in pseudocode, not tested):

```java
import org.gitlab4j.api.GitLabApi.ApiVersion;

    private GitLabApi createGitLabApi(...) {
        for (version : GitLabApi.ApiVersion.versions) {
            GitLabApi gitLabApi =
                    new GitLabApi(version, ...);
            // make sure the server supports this version and loop around otherwise. for example...
            try {
                gitLabApi.getCurrentUser();
                return client;
            } catch (...) {
                // try a lower version
            }
        }   
    }   
```

Yes, to test this you'll need to spin up both a GitLab instance that supports V4 and one that supports V3, since as you need at least two versions to test with and I don't think V5 has been released yet.

---

Finally there is a _separate_ question of when to remove support for V3 from the Jenkins GitLab plugin (note that the GitLab4J library still supports both V3 and V4). According to the README we no longer promise to support versions of GitLab that are that old, so we could rip out support for V3. That could be done in a _separate_ PR that is targeting the main branch if desired, but it is not related to the current RESTEasy migration. If you want to pursue that unrelated cleanup, please open a separate PR to the main branch that does the following:

- Displays a warning on the configuration page when an unsupported version like V3 is configured (note that to preserve compatibility, we need to maintain V3 as a configuration option forever).
- Displays an error if someone has V3 configured and actually tries to do a build with it.

src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnection.java
```diff
@@ -185,21 +150,13 @@ private String getApiToken(String apiTokenId, Item item) {
...
-                    ignoreCertificateErrors,
-                    connectionTimeout,
-                    readTimeout);
+        if (gitLabApi == null) {
+            return new GitLabConnection(name, url, apiTokenId, ignoreCertificateErrors, connectionTimeout, readTimeout);
         }
-        if (clientCache == null) {
-            clientCache = new HashMap<>();
+        if (gitLabApiCache == null) {
+            gitLabApiCache = new HashMap<>();
```
basil: As described above, we cannot break compatibility nor is it desirable to drop support for autodetection. This should be reverted.

src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnection.java
```diff
@@ -277,28 +234,23 @@ public FormValidation doCheckReadTimeout(@QueryParameter Integer value) {
...
             try {
-                new GitLabConnection(
-                                "",
-                                url,
-                                apiTokenId,
-                                clientBuilderId,
-                                ignoreCertificateErrors,
-                                connectionTimeout,
-                                readTimeout)
-                        .getClient(null, null)
```
basil: As described above, we cannot break compatibility nor is it desirable to drop support for autodetection. This should be reverted.

src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnection.java
```diff
@@ -277,28 +234,23 @@ public FormValidation doCheckReadTimeout(@QueryParameter Integer value) {
...
+                new GitLabConnection("", url, apiTokenId, ignoreCertificateErrors, connectionTimeout, readTimeout)
+                        .getGitLabApi(null, null)
+                        .getUserApi()
                         .getCurrentUser();
                 return FormValidation.ok(Messages.connection_success());
             } catch (WebApplicationException e) {
                 return FormValidation.error(Messages.connection_error(e.getMessage()));
             } catch (ProcessingException e) {
                 return FormValidation.error(
                         Messages.connection_error(e.getCause().getMessage()));
```
basil: I don't think we need to catch `ProcessingException` anymore as this could only be thrown from RESTEasy I think. Please address this uniformly throughout the codebase by removing any import of `ProcessingException` from any Java file in `src/main`.

src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnection.java
```diff
@@ -316,14 +268,5 @@ public ListBoxModel doFillApiTokenIdItems(@QueryParameter String url, @QueryPara
...
         }
-
-        public ListBoxModel doFillClientBuilderIdItems() {
-            ListBoxModel model = new ListBoxModel();
-            for (GitLabClientBuilder builder : getAllGitLabClientBuilders()) {
-                model.add(builder.id());
-            }
-
-            return model;
-        }
```
basil: As described above, we cannot break compatibility nor is it desirable to drop support for autodetection. This should be reverted.

src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnectionProperty.java
```diff
@@ -167,18 +168,20 @@ public FormValidation doTestConnection(
...
-                        .getClient(item, jobCredentialId)
+                        .getGitLabApi(item, jobCredentialId)
+                        .getUserApi()
                         .getCurrentUser();
                 return FormValidation.ok(Messages.connection_success());
             } catch (WebApplicationException e) {
                 return FormValidation.error(Messages.connection_error(e.getMessage()));
             } catch (ProcessingException e) {
                 return FormValidation.error(
                         Messages.connection_error(e.getCause().getMessage()));
```
basil: Ditto regarding catching `ProcessingException` being unnecessary since this could only be thrown by `RESTEasy`

src/main/java/com/dabsquared/gitlabjenkins/service/GitLabProjectBranchesService.java
```diff
@@ -34,9 +35,9 @@ public static GitLabProjectBranchesService instance() {
         return gitLabProjectBranchesService;
     }
 
-    public List<String> getBranches(GitLabClient client, String sourceRepositoryString) {
+    public List<String> getBranches(GitLabApi gitLabApi, String sourceRepositoryString) {
```
basil: Here and elsewhere it is creating a lot of diff churn to rename both the type from GitLabClient to GitLabApi and the variable from client to gitLabApi. Why don't you leave the variable named `client` to cut down on diff churn. It is not inaccurate: `GitLabApi` _is_ an API client. And this will cut down on the size of an already very large diff.

src/main/java/com/dabsquared/gitlabjenkins/service/GitLabProjectBranchesService.java
```diff
@@ -47,23 +48,27 @@ public static class BranchLoadingException extends RuntimeException {
...
             }
-            for (Branch branch : client.getBranches(projectId)) {
-                result.add(branch.getName());
+            try {
+                for (Branch branch : gitLabApi.getRepositoryApi().getBranches(projectId)) {
+                    result.add(branch.getName());
+                }
+            } catch (GitLabApiException e) {
+                LOGGER.log(Level.SEVERE, "failed to load branches from repository " + e.getMessage());
             }
```
basil, line 72: I suspect this exception was being propagated rather than logged if there was no log statement in the old code. If true, then the new code should also propagate rather than log.

README.md
```diff
@@ -590,6 +584,8 @@ This saves time in projects where builds can stay long time in a build queue and
 
 ## Compatibility
 
+During the migration of the GitLab Plugin from RESTEasy library to GitLab4J-API as GitLab has dropped support for API V3, the support for GitLab API V3 has also been removed from the plugin and only the latest V4 is supported.
+
```
basil, line 589: I don't see how this has anything to do with the migration from RESTEasy to GitLab4j. GitLab4j supports both V4 and V3 (even version 6.0.0.rc1). The RESTEasy implementation supports both V4 and V3. Dropping support for V3 on the grounds that GitLab < 11 is too old is a completely valid thing to do, but it has nothing to do with the project to migrate to GitLab4j. If doing this is desired, it should be done in a _separate_ PR targeting the main branch as described in https://github.com/jenkinsci/gitlab-plugin/pull/1491/files#r1210826532.

basil:
I reviewed the design but stopped reading after 3-4 source files. On a design level the current effort appears to drop support for autodetection, which I think we want to retain, not only because existing users might have it enabled but also because I think they have a legitimate reason to keep it enabled: in the future, there could exist a GitLab X that supports V5 and V4 and a GitLab X-1 that only supports V4, and users of both may be using this Jenkins plugin, in which case the existing autodetection feature would be desirable. In short I don't see a reason to rip it out, and it doesn't seem prohibitively difficult to retain, so I think we should retain it. I left some more details about how to retain it above.

The other major problem I can see from the above is that it conflates a separate change, which is to drop support for API v3 from this Jenkins plugin (even though it is still supported in GitLab4j). This is a reasonable thing to do, but it is an orthogonal change to this one. As described above, it can be proposed in a separate PR to the main branch. Note that to preserve the ability to read existing configuration files, we need to retain some notion of V3 support on the configuration screen (at least to be able to load old data that is already configured to use V3), even if we don't allow any new builds to take place with V3.

PR #1493, opened 2023-05-30
Lower barrier on project branch for tests (temporarily)

basil:
Based on https://github.com/jenkinsci/gitlab-plugin/pull/1491 it seems clear that it is not just too hard to get the tests to pass in early iterations of this code but it is too hard to get them to even compile. So lower the barrier to a passing build on the project branch from the status quo (compilation of `src/main` and `src/test` but not execution of tests) to just compilation of `src/main` (no compilation or execution of tests). Note that this is temporary. The idea is that once `src/main` is in better shape, a future phase of the project would undo this and get the tests compiling (first) and then executing successfully (after that) as a prerequisite to merging this project to the main branch. Note that this still preserves `InjectedTest` for the Jenkins build to pass, since the CI job needs at least one test in order to mark the overall build as a pass.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,8 +10,5 @@ buildPlugin(
   configurations: [
     [platform: 'linux',   jdk: '11'], // Linux first for coverage report on ci.jenkins.io
     [platform: 'windows', jdk: '17'],
-  ],
-  tests: [
-    skip: true // skip tests
   ]
 )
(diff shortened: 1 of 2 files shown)


basil:
Replayed the build with the `Jenkinsfile` changes here: https://ci.jenkins.io/job/Plugins/job/gitlab-plugin/job/PR-1493/3

basil:
https://ci.jenkins.io/job/Plugins/job/gitlab-plugin/job/PR-1493/3/ worked as expected: did not compile or run any of our own tests, but did run `InjectedTest` as a basic sanity check.

MarkEWaite (approved):

PR #1501, opened 2023-06-09
[GSoC Milestone 1] Migrate resource calls from RESTEasy to GitLab4J-API

harsh-ps-2003:
This PR migrates the GitLab Plugin's resource calls from RESTEasy to GitLab4J-API without affecting the older Webhook implementation. The Proxy Configuration is currently non-functional.

An attempt has been made to adapt Docker Based Functional Tests accordingly.


krisstern:
@basil I have re-enabled the tests and everything seems to be in working order for this PR

krisstern:
## Review Checklist

- [ ] ~~Make sure unit tests are passing~~
- [ ] ~~Make sure integration tests are passing~~
- [x] Review logic of code by module
- [x] Review logic of mapping between `RESTEasy` implementation and `gitlab4j-api` implementation
- [x] Review quality of PR

To be signed off by @krisstern.

src/main/java/com/dabsquared/gitlabjenkins/cause/CauseData.java
```diff
@@ -421,25 +420,6 @@ public String getMergeRequestAssignee() {
         return mergeRequestAssignee;
     }
 
-    @Exported
```
basil, line 424: `@ExportedBean` usually indicates that something is visible in the XML/JSON API, so you'll need to do some research with the old code to see where this exported data was being exposed in the XML/JSON API and ensure that the same data is exposed in the XML/JSON API after this PR.

src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnection.java
```diff
@@ -292,13 +292,13 @@ public FormValidation doTestConnection(
...
+                        .getUserApi()
                         .getCurrentUser();
                 return FormValidation.ok(Messages.connection_success());
             } catch (WebApplicationException e) {
                 return FormValidation.error(Messages.connection_error(e.getMessage()));
-            } catch (ProcessingException e) {
-                return FormValidation.error(
-                        Messages.connection_error(e.getCause().getMessage()));
+            } catch (GitLabApiException e) {
+                return FormValidation.error(Messages.connection_error(e.getMessage()));
```
basil, line 297: How was this tested? Can you show an example of what this error message looks like after an unsuccessful authentication attempt?

src/main/java/com/dabsquared/gitlabjenkins/gitlab/hook/model/PipelineHook.java
```diff
@@ -1,3 +1,6 @@
+/* Note to Reviewers :
+ * Will be removed.
+ */
```
basil, line 3:
```suggestion
/*
 * TODO remove this class because <insert reason here>
 */
```

src/main/java/com/dabsquared/gitlabjenkins/publisher/GitLabVotePublisher.java
```diff
@@ -16,6 +13,10 @@
 import javax.ws.rs.NotFoundException;
 import javax.ws.rs.ProcessingException;
```
basil: Why is `javax.ws.rs.ProcessingException` still being imported? This exception is only thrown by RESTEasy, not GitLab4J.

src/main/java/com/dabsquared/gitlabjenkins/workflow/AcceptGitLabMergeRequestStep.java
```diff
@@ -16,6 +14,9 @@
 import javax.ws.rs.ProcessingException;
```
basil: Why is `javax.ws.rs.ProcessingException` still being imported? This exception is only thrown by RESTEasy, not GitLab4J.

src/main/java/com/dabsquared/gitlabjenkins/workflow/AddGitLabMergeRequestCommentStep.java
```diff
@@ -20,6 +18,9 @@
 import javax.ws.rs.WebApplicationException;
```
basil: Why is `javax.ws.rs.ProcessingException` still being imported? This exception is only thrown by RESTEasy, not GitLab4J.

src/test/java/com/dabsquared/gitlabjenkins/connection/GitLabConnectionConfigSSLTest.java
```diff
@@ -170,7 +170,7 @@ public void doCheckConnection_certificateError() throws IOException {
                 (DescriptorImpl) jenkins.jenkins.getDescriptor(GitLabConnection.class);
 
         FormValidation formValidation =
-                descriptor.doTestConnection("https://localhost:" + port + "/gitlab", API_TOKEN_ID, "v3", false, 10, 10);
+                descriptor.doTestConnection("https://localhost:" + port + "/gitlab", API_TOKEN_ID, "v4", false, 10, 10);
```
basil: How is this change specifically required for the migration from RESTEasy to GitLab4J? If it is not required for the migration but is rather a generally desirable cleanup, please remove this hunk from this PR and submit it to the main branch in a separate pull request.

src/main/java/com/dabsquared/gitlabjenkins/gitlab/api/impl/AutodetectGitLabClientBuilder.java
```diff
@@ -18,11 +18,25 @@ public AutodetectGitLabClientBuilder() {
 
     @Override
     @NonNull
-    public GitLabClient buildClient(
+    public GitLabApi buildClient(
```
basil, line 25: The original version of this code was more elegant I think. It retrieved the sorted list of all builders (autodetecting, V3, and V4), removed itself (the autodetecting builder) from the list, and then delegated to the existing implementation in the V3 and V4 builders. Your version duplicates the implementation in the autodetecting builder. It would be more elegant to reuse the existing code in the V3 and V4 builders. This can be done in this method with something like this:

```java
    @Override
    @NonNull
    public GitLabApi buildClient(String url, String token, boolean ignoreCertificateErrors, int connectionTimeout, int readTimeout) {
        Collection<GitLabClientBuilder> candidates = new ArrayList<>(getAllGitLabClientBuilders());
        candidates.remove(this);
        return autodetectOrDie();
    }

    @NonNull
    private GitLabApi autodetectOrDie(Collection<GitLabClientBuilder> candidates) {
        GitLabApi client = autodetect(candidates);
        if (client != null) {
            return client;
        }

        throw new NoSuchElementException("no client-builder found that supports server at " + url);
    }

    @CheckForNull
    private GitLabApi autodetect(Collection<GitLabClientBuilder> candidates) {
        for (GitLabClientBuilder candidate : candidates) {
            GitLabApi client =
                    candidate.buildClient(url, token, ignoreCertificateErrors, connectionTimeout, readTimeout);
            try {
                client.getCurrentUser();
                return client;
            } catch (<whatever type getCurrentUser would throw if the user was not found> ignored) {
                // api-endpoint not found (== api-level not supported by this client)
            }
        }

        return null;
    }
```

src/main/java/com/dabsquared/gitlabjenkins/gitlab/api/impl/V3GitLabClientBuilder.java
```diff
@@ -1,24 +1,33 @@
...
+import org.gitlab4j.api.GitLabApi;
+import org.gitlab4j.api.GitLabApi.ApiVersion;
+import org.gitlab4j.api.GitLabApiException;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 
 @Extension
 @Restricted(NoExternalUse.class)
-public final class V3GitLabClientBuilder extends ResteasyGitLabClientBuilder {
-    private static final int ORDINAL = 2;
```
basil, line 12: Why remove this constant? If there is no reason to remove it then please keep it.

src/main/java/com/dabsquared/gitlabjenkins/gitlab/api/impl/V3GitLabClientBuilder.java
```diff
@@ -1,24 +1,33 @@
...
 
     public V3GitLabClientBuilder() {
-        super(V3GitLabApiProxy.ID, ORDINAL, V3GitLabApiProxy.class, MERGE_REQUEST_ID_PROVIDER);
+        super("V3", 2);
+    }
+
+    @Override
+    public GitLabApi buildClient(
+            String url, String token, boolean ignoreCertificateErrors, int connectionTimeout, int readTimeout) {
+        GitLabApi client = null;
```
basil: Redundant initialization.

src/main/java/com/dabsquared/gitlabjenkins/gitlab/api/impl/V4GitLabClientBuilder.java
```diff
@@ -1,24 +1,33 @@
...
 
     public V4GitLabClientBuilder() {
-        super(V4GitLabApiProxy.ID, ORDINAL, V4GitLabApiProxy.class, MERGE_REQUEST_ID_PROVIDER);
+        super("V4", 1);
+    }
+
+    @Override
+    public GitLabApi buildClient(
+            String url, String token, boolean ignoreCertificateErrors, int connectionTimeout, int readTimeout) {
+        GitLabApi client = null;
```
basil: Redundant initialization.

src/main/java/com/dabsquared/gitlabjenkins/gitlab/api/impl/V4GitLabClientBuilder.java
```diff
@@ -1,24 +1,33 @@
...
+        try {
+            client = new GitLabApi(ApiVersion.V4, url, token);
+            client.getUserApi().getCurrentUser();
+            client.setIgnoreCertificateErrors(ignoreCertificateErrors);
+            client.setRequestTimeout(connectionTimeout, readTimeout);
+            return client;
+        } catch (GitLabApiException e) {
+            return null;
```
basil: The abstract `com.dabsquared.gitlabjenkins.gitlab.api.GitLabClientBuilder#buildClient` method is annotated with `@NonNull`, so this is in violation of the API contract. SpotBugs correctly raises a warning.

src/main/java/com/dabsquared/gitlabjenkins/gitlab/api/impl/V3GitLabClientBuilder.java
```diff
@@ -1,24 +1,33 @@
...
+        try {
+            client = new GitLabApi(ApiVersion.V3, url, token);
+            client.getUserApi().getCurrentUser();
+            client.setIgnoreCertificateErrors(ignoreCertificateErrors);
+            client.setRequestTimeout(connectionTimeout, readTimeout);
+            return client;
+        } catch (GitLabApiException e) {
+            return null;
```
basil: The abstract `com.dabsquared.gitlabjenkins.gitlab.api.GitLabClientBuilder#buildClient` method is annotated with `@NonNull`, so this is in violation of the API contract. SpotBugs correctly raises a warning.

src/spotbugs/excludesFilter.xml
```diff
@@ -138,4 +138,10 @@
   <Match>
     <Bug pattern="RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT" />
   </Match>
+  <Match>
+    <Bug pattern="NP_NULL_ON_SOME_PATH" />
+  </Match>
+  <Match>
+    <Bug pattern="NP_NONNULL_RETURN_VIOLATION" />
+  </Match>
```
basil: Why are these suppressions being added? They look like legitimate warnings to me.

pom.xml
```diff
@@ -84,6 +84,10 @@
       <groupId>io.jenkins.plugins</groupId>
       <artifactId>caffeine-api</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.jenkins.plugins</groupId>
+      <artifactId>gitlab-api</artifactId>
+    </dependency>
```
basil: In addition to the addition of this dependency we also need to remove RESTEasy and any imports of `javax.ws.rs.*`. Otherwise we retain dead code that is no longer needed as well as a potential source of security vulnerabilities.

basil (changes requested):
Significant progress in the right direction. Nice job!

Also some significant changes that still need to be made. See review below.

src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnection.java
harsh-ps-2003: Screenshot 2023-06-25 at 11 04 57 PM
Screenshot 2023-06-25 at 11 09 18 PM
@basil On reproducing the exception I get InvocationTargetException which will be wrapped in GitLabApiExcetion.

src/spotbugs/excludesFilter.xml
krisstern: I think there were some problems with some modules before, but maybe we should target specific modules if we really do want to add these exlusions. @harsh-ps-2003 Would you like to follow up?

src/main/java/com/dabsquared/gitlabjenkins/cause/CauseData.java
harsh-ps-2003: Actually this change is temporary. Will be reverted back in a similar fashion in Milestone 2 PR.

src/spotbugs/excludesFilter.xml
harsh-ps-2003: I don't think these suppressions are required. I don't see any errors when doing `mvn verify` without them.

krisstern: Yes, I can confirm now that these changes are no longer required and have been reverted.

src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnection.java
basil: Any time Angry Jenkins shows up in the UI there is a bug.

src/main/java/com/dabsquared/gitlabjenkins/cause/CauseData.java
basil: Why is it temporary?

src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnection.java
harsh-ps-2003: Oh! I was unaware and thought this was normal behaviour. I didn't realise that this thing can create a bug. Sorry! will look into it.

src/main/java/com/dabsquared/gitlabjenkins/cause/CauseData.java
harsh-ps-2003: Because when we will break the CauseData according to different events recieved we will have the access to GitLabApi intance in the respective causedata's, thus allowing us to get the Merge Requests. But in the current implimentation we did not have the GitLabApi intance directly availabe in the CauseData, so i shifted the getMergeRequest method to wherever it was required.

src/test/java/com/dabsquared/gitlabjenkins/publisher/GitLabCommitStatusPublisherTest.java
```diff
@@ -99,23 +99,23 @@ public void matrixAggregatable() throws InterruptedException, IOException {
     @Test
     public void running_v3() throws UnsupportedEncodingException {
         AbstractBuild build = mockBuild(GITLAB_CONNECTION_V3, null, "test/project.git");
-        HttpRequest[] requests = prepareCheckCommitAndUpdateStatusRequests("v3", build, BuildState.running);
+        HttpRequest[] requests = prepareCheckCommitAndUpdateStatusRequests("v3", build, CommitBuildState.RUNNING);
```
krisstern:
```suggestion
        HttpRequest[] requests = prepareCheckCommitAndUpdateStatusRequests("V3", build, CommitBuildState.RUNNING);
```

src/test/java/com/dabsquared/gitlabjenkins/publisher/GitLabCommitStatusPublisherTest.java
```diff
@@ -125,8 +125,9 @@ public void runningWithDotInProjectId() throws IOException {
         AbstractBuild build = mockBuild(GITLAB_CONNECTION_V4, null, "test/project.test.git");
         HttpRequest[] requests = new HttpRequest[] {
             prepareGetProjectResponse("test/project.test"),
-            prepareExistsCommitWithSuccessResponse("v4", String.valueOf(PROJECT_ID)),
-            prepareUpdateCommitStatusWithSuccessResponse("v4", String.valueOf(PROJECT_ID), build, BuildState.running)
+            prepareExistsCommitWithSuccessResponse("V4", String.valueOf(PROJECT_ID)),
+            prepareUpdateCommitStatusWithSuccessResponse(
+                    "v4", String.valueOf(PROJECT_ID), build, CommitBuildState.RUNNING)
```
krisstern:
```suggestion
                    "V4", String.valueOf(PROJECT_ID), build, CommitBuildState.RUNNING)
```

src/test/java/com/dabsquared/gitlabjenkins/publisher/GitLabMessagePublisherTest.java
```diff
@@ -235,7 +236,7 @@ public void successWithCustomNote() throws IOException, InterruptedException {
                 false,
                 false,
                 false,
-                prepareSendMessageWithSuccessResponse("v4", MERGE_REQUEST_IID, defaultNote));
+                prepareSendMessageWithSuccessResponse("4", MERGE_REQUEST_IID, defaultNote));
```
krisstern:
```suggestion
                prepareSendMessageWithSuccessResponse("V4", MERGE_REQUEST_IID, defaultNote));
```

krisstern (changes requested):

krisstern:
@harsh-ps-2003 I have made some requests for changes. I will need to test interactively once again after these changes have been made, and can merge the PR after that.

krisstern (approved):
Thanks for your hard work @harsh-ps-2003! I tried it interactively with a push event and it worked. For testing for merge request I am seeing the below: 

Screenshot 2023-07-30 at 4 27 17 PM

Which may mean that I need a bit working on my own setup.

krisstern:
> Thanks for your hard work @harsh-ps-2003! I tried it interactively with a push event and it worked. For testing for merge request I am seeing the below:
> 
> Screenshot 2023-07-30 at 4 27 17 PM
> Which may mean that I need a bit working on my own setup.

So to follow up, once I have set up a proper merge request for the merge test the test was passed:

Screenshot 2023-07-30 at 5 55 19 PM

PR #1510, opened 2023-06-19
Remove outdated version reference

MarkEWaite:
## Remove outdated version reference

Better to not mention a specific version than to mention a specific version that is many years outdated.


gounthar (approved):
Good idea!

PR #1549, opened 2023-07-19
do not sign git commits in tests

car-roll:
Remove signing git commits in tests.
I had been running into test failures regarding git commit signing that I believe were related to my local git environment. Signing the commits did not seem necessary regarding tests, so I thought removing them could reduce some problems and not affect testing too much.


julieheard (approved):

MarkEWaite (approved):
@dependabot squash and merge

PR #1594, opened 2023-11-21
Added support for retrieving merge request ref

olpal:
Added support for merge MERGE enum type when determining the branch name or tag. Without this, a duplicate pipeline is generated in Gitlab when using a merge request pipeline from the gitlab side.


olpal:
This should not be merged

PR #1603, opened 2023-12-30
Replace wiki link with link to plugins site

MarkEWaite:
## Replace wiki link with link to plugins site

The wiki link already redirects to the plugins site.  Rather than rely on the redirection, link directly to the final destination URL.


krisstern:
Thanks @MarkEWaite!

PR #1604, opened 2023-12-30
Rely on default Jenkins PR template

MarkEWaite:
## Rely on default Jenkins PR template

Custom template does not mention several important topics.  For example, the default template guides the submitter to use a branch that is not the master branch.  The default template encourages the submitter to provide tests. The default template encourages the submitter to describe the testing that was performed on the pull request.  Those are all good things that should be used for the GitLab plugin.

https://github.com/jenkinsci/.github/blob/master/.github/pull_request_template.md is the default template.


krisstern (approved):
LGTM

PR #1613, opened 2024-01-13
Fix null check condition in PipelineHookTriggerHandlerImpl.java

shivajee98:
Fix null check condition in PipelineHookTriggerHandlerImpl.java

The condition `hook.getRepository() == null || hook.getRepository() == null` is redundant.
This commit removes the duplicate check, leaving only `hook.getRepository() == null`.
This fix ensures correct null checking for the `getRepository()` method in PipelineHookTriggerHandlerImpl.java.


MarkEWaite (approved):
Thanks!  Nice find.

PR #1614, opened 2024-01-14
Replaced Arguements With Empty Array

shivajee98:


krisstern:
Hi @shivajee98,
What is the motivation behind this PR and does it resolve a particular issue in the issue tracker?

shivajee98:
Using new String[0] is considered more idiomatic in modern java development
as  the array is then dynamically resized as needed.
My motivation was to follow a more modern and concise coding style.

krisstern:
I am guessing you mean something like the one reference here https://stackoverflow.com/a/65902425/9959070:

> There are two styles to convert a collection to an array: either using a pre-sized array (like c.toArray(new String[c.size()])) or using an empty array (like c.toArray(new String[0]). In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow. However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version. Also passing pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray call which may result in extra nulls at the end of the array, if the collection was concurrently shrunk during the operation. This inspection allows to follow the uniform style: either using an empty array (which is recommended in modern Java) or using a pre-sized array (which might be faster in older Java versions or non-HotSpot based JVMs).

That makes sense, but it would be great for you to provide some context for the PR instead of having an empty description. We do not encourge that practice in open-source software development, as we would have no idea why the contributor is proposing the change.

krisstern (approved):
LGTM

krisstern:
Thanks!

shivajee98:
> I am guessing you mean something like the one reference here https://stackoverflow.com/a/65902425/9959070:
> 
> > There are two styles to convert a collection to an array: either using a pre-sized array (like c.toArray(new String[c.size()])) or using an empty array (like c.toArray(new String[0]). In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow. However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version. Also passing pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray call which may result in extra nulls at the end of the array, if the collection was concurrently shrunk during the operation. This inspection allows to follow the uniform style: either using an empty array (which is recommended in modern Java) or using a pre-sized array (which might be faster in older Java versions or non-HotSpot based JVMs).
> 
> That makes sense, but it would be great for you to provide some context for the PR instead of having an empty description. We do not encourge that practice in open-source software development, as we would have no idea why the contributor is proposing the change.

Thank you for your feedback. I appreciate the guidance on providing meaningful context for the PR. I apologize for the oversight in this instance.

Moving forward, I assure you that I will make it a priority to offer detailed descriptions for my contributions. I understand the importance of clear communication in open-source development, and I am committed to providing comprehensive explanations for any changes I propose.
Thank you for your understanding.

krisstern:
Hi @shivajee98, I am afraid we may need to revert the changes due to the concern that an empty Array will stay with dimension 0 after initialisation. This goes against what we are trying to achieve as Array is unlike ArrayList and is of a fixed size.

PR #1618, opened 2024-01-15
Refactor GitLabConnectionProperty code

shivajee98:
Replace explicit cast and null check with Objects.requireNonNull() for improved robustness and readability. 

- Replaced: GitLabConnectionConfig descriptor = (GitLabConnectionConfig) Jenkins.getInstance().getDescriptor(GitLabConnectionConfig.class);
- With: GitLabConnectionConfig descriptor = Objects.requireNonNull(Jenkins.getInstance()).getDescriptor(GitLabConnectionConfig.class);

This change enhances the code by providing a more explicit null check, ensuring that the descriptor variable is guaranteed to be non-null. The use of Objects.requireNonNull() makes the code clearer and more modern.

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnectionProperty.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnectionProperty.java
@@ -30,8 +30,6 @@ import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerRequest;
 import org.kohsuke.stapler.interceptor.RequirePOST;
 
-import java.util.Objects;
-
 /**
  * @author Robin Müller
  */
@@ -88,7 +86,9 @@ public class GitLabConnectionProperty extends JobProperty<Job<?, ?>> {
         Job<?, ?> job = build.getParent();
         if (job != null) {
             final GitLabConnectionProperty connectionProperty = job.getProperty(GitLabConnectionProperty.class);
-            if (connectionProperty != null) return connectionProperty.getClient();
+            if (connectionProperty != null) {
+                return connectionProperty.getClient();
+            }
         }
         return null;
     }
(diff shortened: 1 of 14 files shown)


src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnectionProperty.java
```diff
@@ -109,13 +110,15 @@ public boolean isApplicable(Class<? extends Job> jobType) {
...
             return req.bindJSON(GitLabConnectionProperty.class, formData);
         }
 
         public ListBoxModel doFillGitLabConnectionItems() {
             ListBoxModel options = new ListBoxModel();
-            GitLabConnectionConfig descriptor =
-                    (GitLabConnectionConfig) Jenkins.getInstance().getDescriptor(GitLabConnectionConfig.class);
+            GitLabConnectionConfig descriptor = (GitLabConnectionConfig)
+                    Objects.requireNonNull(Jenkins.getInstance()).getDescriptor(GitLabConnectionConfig.class);
+            assert descriptor != null;
```
krisstern, line 121: Can we use the if-else statement instead of assert here?

src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnectionProperty.java
```diff
@@ -153,8 +156,9 @@ public FormValidation doTestConnection(
             item.checkPermission(Item.CONFIGURE);
             try {
                 GitLabConnection gitLabConnectionTested = null;
-                GitLabConnectionConfig descriptor =
-                        (GitLabConnectionConfig) Jenkins.getInstance().getDescriptor(GitLabConnectionConfig.class);
+                GitLabConnectionConfig descriptor = (GitLabConnectionConfig)
+                        Objects.requireNonNull(Jenkins.getInstance()).getDescriptor(GitLabConnectionConfig.class);
+                assert descriptor != null;
```
krisstern, line 161: Again, same as above: Can we use the if-else statement instead of assert here?

krisstern (changes requested):

src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnectionProperty.java
shivajee98: Certainly, I'd be happy to provide more context on why I chose to use assert instead of an if-else statement in that particular scenario.

1. Intent:
   - assert statements are primarily used for debugging and catching programming errors during development. They express assumptions about the state of the program.

2. Preconditions and Fail Fast:
   - I used assert to check critical preconditions – conditions that are crucial for the correct functioning of the code. This aligns with the "fail fast" principle, helping us detect and address errors as early as possible in the development process.

3. Reducing Cognitive Load:
   - Using assert for debugging-related checks separates assumptions from regular control flow logic, making the code more readable. It also emphasizes the significance of critical conditions.

krisstern: These code snippets are not for testing purposes, but is used instead for program flow control. Can you please update these?

shivajee98: ok sure

krisstern: For a context of what I am saying please refer to https://www.geeksforgeeks.org/assertions-in-java/ where it says:
> It is mainly used for testing purposes during development.

PR #1622, opened 2024-01-16
Revert "Replaced Arguements With Empty Array"

krisstern:
Reverts jenkinsci/gitlab-plugin#1614


AniketNS:
Hey @krisstern, Can I ask you why did you revert this commit?

krisstern:
Hi @AniketNS, as I have explained in https://github.com/jenkinsci/gitlab-plugin/pull/1614#issuecomment-1894096472, a Java Array is always of fixed size and not of dynamic size. If you fixed its length to be 0 then its sole purpose will be to serve as a zero-length array, which in most cases serve as a placeholder. In our context we want the Arrays to be of a certain length to begin with, hence the revert of the PR.

AniketNS:
Ohh ok. Thanks for the detailed explanation @krisstern

krisstern:
No problem

PR #1646, opened 2024-03-23
Improve commit message creation

AniketNS:
The commit message generation is refactored for better readability and efficiency in `AcceptGitLabMergeRequestStep.java` file. Made some changes in getCommitMessage() method.

### Testing done

```[tasklist]
### Submitter checklist
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Please describe what you did
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue
```

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/workflow/AcceptGitLabMergeRequestStep.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/workflow/AcceptGitLabMergeRequestStep.java
@@ -122,14 +122,13 @@ public class AcceptGitLabMergeRequestStep extends Step {
         }
 
         private String getCommitMessage(MergeRequest mergeRequest) {
-            if (!step.useMRDescription) return step.mergeCommitMessage;
+            if (!step.useMRDescription || StringUtils.isEmpty(step.mergeCommitMessage)) {
+                return step.mergeCommitMessage;
+            }
 
-            String message = "Merge branch '" + mergeRequest.getSourceBranch() + "' into '"
-                    + mergeRequest.getTargetBranch() + "'\n\n" + mergeRequest.getTitle()
-                    + "\n\n" + mergeRequest.getDescription()
-                    + "\n\n" + "See merge request !"
-                    + mergeRequest.getIid();
-            return message;
+            return String.format("Merge branch '%s' into '%s'\n\n%s\n\n%s\n\nSee merge request !%d",
+                    mergeRequest.getSourceBranch(), mergeRequest.getTargetBranch(),
+                    mergeRequest.getTitle(), mergeRequest.getDescription(), mergeRequest.getIid());
         }
 
         private void println(String message) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/dabsquared/gitlabjenkins/workflow/AcceptGitLabMergeRequestStep.java
```diff
@@ -122,14 +122,13 @@ protected Void run() throws Exception {
...
-
-            String message = "Merge branch '" + mergeRequest.getSourceBranch() + "' into '"
-                    + mergeRequest.getTargetBranch() + "'\n\n" + mergeRequest.getTitle()
-                    + "\n\n" + mergeRequest.getDescription()
-                    + "\n\n" + "See merge request !"
-                    + mergeRequest.getIid();
-            return message;
+            if (!step.useMRDescription || StringUtils.isEmpty(step.mergeCommitMessage)) {
+                return step.mergeCommitMessage;
+            }
```
MarkEWaite: Please explain why you changed the logic to return the empty string if the `mergeCommitMessage` is an empty string.  That seems like a behavior change and I don't see anything in the commit message or the pull request description that explains why the behavior has been changed.
```suggestion
            if (!step.useMRDescription) return step.mergeCommitMessage;
```

src/main/java/com/dabsquared/gitlabjenkins/workflow/AcceptGitLabMergeRequestStep.java
```diff
@@ -122,14 +122,13 @@ protected Void run() throws Exception {
...
+
+            return String.format("Merge branch '%s' into '%s'\n\n%s\n\n%s\n\nSee merge request !%d",
```
MarkEWaite: As far as I can tell from reading the Javadoc, for [format strings](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Formatter.html#syntax), the newline characters need to be replaced with `%n`
```suggestion
            return String.format("Merge branch '%s' into '%s'%n%n%s%n%n%s%n%nSee merge request !%d",
```

MarkEWaite (changes requested):
Thanks for the pull request.  The pull request appears to have changed the logic.  If that was intentional, then a comment is needed in the commit or in the pull request description to explain the change.

src/main/java/com/dabsquared/gitlabjenkins/workflow/AcceptGitLabMergeRequestStep.java
AniketNS: Yes @MarkEWaite, It was intentional. when we check the `!step.useMRDescription` condition, if this is true then it return the `step.mergeCommitMessage`. And in this case the Commit Message can be empty, so it can lead to a null pointer exception IMO. That's why I added this check. Correct me if I'm wrong.

src/main/java/com/dabsquared/gitlabjenkins/workflow/AcceptGitLabMergeRequestStep.java
AniketNS: Yeah, you're indeed correct. `%n` is good choice here. Also it is platform independent and this would be the correct approach for Java's String.format() method. 

Thanks for the clarification. I'll make the changes.

MarkEWaite (approved):
Thanks

PR #1711, opened 2024-10-22
Avoid caching the credential, instead get the credential value whenev…

arechavarria:
### Description

This pull request introduces a new `GitlabCredentialResolver` class to manage the retrieval of credentials for GitLab connections more centrally and securely within the Jenkins GitLab plugin. The previous logic for handling API tokens has been refactored out of `GitLabConnection` and related classes into this new resolver class, improving code readability and maintainability. 

Key changes include:
- Creation of the `GitlabCredentialResolver` class to handle credential resolution based on `Item` context and credential IDs.
- Refactoring `GitLabClientBuilder` and its implementations to accept a `GitlabCredentialResolver` instead of a plain API token string.
- Modifying `ApiHeaderTokenFilter` to utilize `GitlabCredentialResolver` for injecting API tokens into HTTP requests.
- Updates to unit tests to accommodate the new method of resolving credentials, ensuring compatibility with the new design.

### Testing done

- **Unit Tests**: Updated existing tests to use `GitlabCredentialResolver`, ensuring they pass with the refactored code. This includes modifying tests like `AutodetectingGitLabClientTest` and `TestUtility`.
- **Manual Testing**: Verified the following scenarios manually:
  - Connection to GitLab using both global API tokens and job-specific credentials.
  - Ensured that the new `GitlabCredentialResolver` properly tracks credentials when associated with specific Jenkins items.
  - Verified that API requests include the correct `PRIVATE_TOKEN` header for both global and job-specific tokens.
- **Automated Test Execution**: Ran all existing tests to ensure that there are no regressions introduced by this refactoring.

### Submitter checklist
- [x] Link to relevant issues in GitHub or Jira
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnection.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnection.java
@@ -1,10 +1,5 @@
 package com.dabsquared.gitlabjenkins.connection;
 
-import static com.cloudbees.plugins.credentials.CredentialsProvider.lookupCredentials;
-import static com.dabsquared.gitlabjenkins.gitlab.api.GitLabClientBuilder.getAllGitLabClientBuilders;
-import static com.dabsquared.gitlabjenkins.gitlab.api.GitLabClientBuilder.getGitLabClientBuilderById;
-
-import com.cloudbees.plugins.credentials.CredentialsMatchers;
 import com.cloudbees.plugins.credentials.CredentialsProvider;
 import com.cloudbees.plugins.credentials.CredentialsScope;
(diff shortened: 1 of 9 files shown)


krisstern:
I am not sure if this is the approach we would like to adopt. Also, we do not accept a PR opening with a `main` branch from your fork.

arechavarria:
What's not clear? It turns out there is a related issue, when for example aws secret manager is used, there are credentials that are rotated, however the plugin caches the client and the same token is always used, which causes that when the credential rotates the plugin starts to generate errors with Gitlab, because it continues using the same token

PR #1712, opened 2024-10-22
Avoid caching the credential, instead get the credential value whenever it is required through the request filter 

### LINKED ISSUE #963: add ability to flush the cached connection
### Context
- **Gitlab plugin version**: 1.5.12
- **Gitlab version**: GitLab Community Edition 11.10.1
- **Jenkins version**: 2.121.3
- **Job type**: Freestyle (but that should be irrelevant)

Today, I experienced a situation where Jenkins wasn't able to submit build status to GitLab, saying `Failed to update Gitlab commit status for project '19795': HTTP 401 Unauthorized`.

I revoked the access token on GitLab side, created a new token, and performed an in-place update of the token in the Jenkins credentials associated with the GitLab connection I have defined in my Jenkins. The error persisted.

I looked into the GitLab plugin source code (= this repo) and eventually found that if the connection has been created once, it's cached for subsequent use: https://github.com/jenkinsci/gitlab-plugin/blob/master/src/main/java/com/dabsquared/gitlabjenkins/connection/GitLabConnection.java#L48 Using the Jenkins script console, I confirmed that the cached client uses the old token.

### Feature request

I now understand that I should probably have removed the GitLab connection entirely and create a new one, but that can hardly be considered intuitive.

Instead, I'd like to propose a new feature: next to the `Test Connection` button, add another one, called e.g. _Clear Connection Cache_ (perhaps with a help text that includes something like "this is useful when you update the credentials"). This button would set the `apiCache` field to `null`, forcing creation of a new client. (It should also dispose all resources that the client holds, but I don't know anything about that.)

arechavarria:
Fixes #963. 

### Description

This pull request introduces a new `GitlabCredentialResolver` class to manage the retrieval of credentials for GitLab connections more centrally and securely within the Jenkins GitLab plugin. The previous logic for handling API tokens has been refactored out of `GitLabConnection` and related classes into this new resolver class, improving code readability and maintainability.

Key changes include:
- Creation of the `GitlabCredentialResolver` class to handle credential resolution based on `Item` context and credential IDs.
- Refactoring `GitLabClientBuilder` and its implementations to accept a `GitlabCredentialResolver` instead of a plain API token string.
- Modifying `ApiHeaderTokenFilter` to utilize `GitlabCredentialResolver` for injecting API tokens into HTTP requests.
- Updates to unit tests to accommodate the new method of resolving credentials, ensuring compatibility with the new design.
- **Fix for AWS Secret Manager Issue**: Addressed an issue where, if AWS Secret Manager is used with a rotation mechanism, the plugin encounters problems because the client is cached and always uses the token with which it was initially created. This change ensures that credentials are retrieved each time they are used, allowing the plugin to work correctly with rotated tokens.

### Testing done

- **Unit Tests**: Updated existing tests to use `GitlabCredentialResolver`, ensuring they pass with the refactored code. This includes modifying tests like `AutodetectingGitLabClientTest` and `TestUtility`.
- **Manual Testing**: Verified the following scenarios manually:
  - Connection to GitLab using both global API tokens and job-specific credentials.
  - Ensured that the new `GitlabCredentialResolver` properly tracks credentials when associated with specific Jenkins items.
  - Verified that API requests include the correct `PRIVATE_TOKEN` header for both global and job-specific tokens.
  - Tested the resolution of credentials in environments with AWS Secret Manager to confirm that rotated tokens are correctly retrieved and used.
- **Automated Test Execution**: Ran all existing tests to ensure that there are no regressions introduced by this refactoring.

### Submitter checklist
- [x] Link to relevant issues in GitHub or Jira
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/java/com/dabsquared/gitlabjenkins/connection/GitlabCredentialResolver.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/connection/GitlabCredentialResolver.java
@@ -27,6 +27,7 @@ public class GitlabCredentialResolver {
         return credentialsId;
     }
 
+    @Restricted(NoExternalUse.class)
     public void setCredentialsId(String credentialsId) {
         this.credentialsId = credentialsId;
     }
--- a/src/main/java/com/dabsquared/gitlabjenkins/connection/GitlabCredentialResolver.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/connection/GitlabCredentialResolver.java
@@ -20,6 +20,7 @@ public class GitlabCredentialResolver {
         return item;
     }
 
+    @Restricted(NoExternalUse.class)
     public void setItem(Item item) {
         this.item = item;
     }
--- a/src/main/java/com/dabsquared/gitlabjenkins/connection/GitlabCredentialResolver.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/connection/GitlabCredentialResolver.java
@@ -1,6 +1,8 @@
 package com.dabsquared.gitlabjenkins.connection;
 
 import hudson.model.Item;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
 
 public class GitlabCredentialResolver {
 
--- a/src/main/java/com/dabsquared/gitlabjenkins/connection/GitlabCredentialResolver.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/connection/GitlabCredentialResolver.java
@@ -22,6 +22,7 @@ public class GitlabCredentialResolver {
         this.item = item;
     }
 
+    @Restricted(NoExternalUse.class)
     public String getCredentialsId() {
         return credentialsId;
     }
--- a/src/main/java/com/dabsquared/gitlabjenkins/connection/GitlabCredentialResolver.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/connection/GitlabCredentialResolver.java
@@ -7,8 +7,7 @@ public class GitlabCredentialResolver {
     private Item item;
     private String credentialsId;
 
-    public GitlabCredentialResolver() {
-    }
+    public GitlabCredentialResolver() {}
 
     public GitlabCredentialResolver(Item item, String credentialsId) {
         this.item = item;
--- a/src/main/java/com/dabsquared/gitlabjenkins/connection/GitlabCredentialResolver.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/connection/GitlabCredentialResolver.java
@@ -16,6 +16,7 @@ public class GitlabCredentialResolver {
         this.credentialsId = credentialsId;
     }
 
+    @Restricted(NoExternalUse.class)
     public Item getItem() {
         return item;
     }
--- a/src/main/java/com/dabsquared/gitlabjenkins/connection/GitlabCredentialResolver.java
+++ b/src/main/java/com/dabsquared/gitlabjenkins/connection/GitlabCredentialResolver.java
@@ -1,0 +1,33 @@
+package com.dabsquared.gitlabjenkins.connection;
+
+import hudson.model.Item;
+
+public class GitlabCredentialResolver {
+
+    private Item item;
+    private String credentialsId;
+
+    public GitlabCredentialResolver() {
+    }
+
+    public GitlabCredentialResolver(Item item, String credentialsId) {
+        this.item = item;
+        this.credentialsId = credentialsId;
+    }
+
+    public Item getItem() {
+        return item;
+    }
+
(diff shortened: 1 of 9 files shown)


krisstern:
https://github.com/jenkinsci/gitlab-plugin/pull/1711#issuecomment-2428185919:
> What's not clear? It turns out there is a related issue, when for example aws secret manager is used, there are credentials that are rotated, however the plugin caches the client and the same token is always used, which causes that when the credential rotates the plugin starts to generate errors with Gitlab, because it continues using the same token

krisstern:
Hi @arechavarria Thanks for the PR! But where is the item for the "Link to relevant issues in GitHub or Jira" you have indicated in the checklist that there is one?

arechavarria:
> Hi @arechavarria Thanks for the PR! But where is the item for the "Link to relevant issues in GitHub or Jira" you have indicated in the checklist that there is one? 

Hello, with pleasure 

I added the relationship with the issue: 
 https://github.com/jenkinsci/gitlab-plugin/issues/963

krisstern:
Hi @daniel-beck would it be able for you to take a look also and see if this PR would have any security issue(s) if merged as is?

src/main/java/com/dabsquared/gitlabjenkins/connection/GitlabCredentialResolver.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public Item getItem() {
+        return item;
+    }
+
+    public void setItem(Item item) {
+        this.item = item;
+    }
+
+    public String getCredentialsId() {
```
krisstern, line 30:
```suggestion
    @Restricted(NoExternalUse.class)
    public String getCredentialsId() {
```

src/main/java/com/dabsquared/gitlabjenkins/connection/GitlabCredentialResolver.java
```diff
@@ -0,0 +1,32 @@
...
+        return credentialsId;
+    }
+
+    public void setCredentialsId(String credentialsId) {
```
krisstern, line 35:
```suggestion
    @Restricted(NoExternalUse.class)
    public void setCredentialsId(String credentialsId) {
```

krisstern (changes requested):
I have a gut feeling we should be more careful about retricting any external use of the two methods concerned?

src/main/java/com/dabsquared/gitlabjenkins/connection/GitlabCredentialResolver.java
```diff
@@ -0,0 +1,36 @@
...
+    private String credentialsId;
+
+    public GitlabCredentialResolver() {}
+
+    public GitlabCredentialResolver(Item item, String credentialsId) {
+        this.item = item;
+        this.credentialsId = credentialsId;
+    }
+
+    public Item getItem() {
```
krisstern, line 20:
```suggestion
    @Restricted(NoExternalUse.class)
    public Item getItem() {
```

src/main/java/com/dabsquared/gitlabjenkins/connection/GitlabCredentialResolver.java
```diff
@@ -0,0 +1,36 @@
...
+        return item;
+    }
+
+    public void setItem(Item item) {
```
krisstern, line 25:
```suggestion
    @Restricted(NoExternalUse.class)
    public void setItem(Item item) {
```

src/main/java/com/dabsquared/gitlabjenkins/connection/GitlabCredentialResolver.java
krisstern: This is just to be safe, since its usage is wrapped around the already restricted `getApiToken` method in "GitLabClientBuilder.java".

src/main/java/com/dabsquared/gitlabjenkins/connection/GitlabCredentialResolver.java
krisstern: We may need this too because of its invocation in the unrestricted `getClient` method in "GitLabConnection.java".

krisstern (changes requested):

krisstern:
Hi @arechavarria I will let this PR sit here for 24 to 48 hours. If there is no objection from the community within that time frame, I will proceed to merge it.

krisstern (approved):
LGTM

arechavarria:
Thanks @krisstern

arechavarria:
@krisstern I ask you at what point the release is managed so that the plugin is available in the marketplace?

krisstern:
Hi @arechavarria we have another PR #1713 pending to be merged soon. So we will need to wait until afterwards when a new release will be cut.

PR #1775, opened 2025-05-17
Fix docker compose port conflict

dodo920306:
The default jenkins agent port conflicts with the compose file which currently configures Jenkins with a JAVA_TOOL_OPTIONS setting that opens JDWP debugging on port 50000.

This PR changes the JDWP debug port to 50001 to resolve the issues, and port 50000 still remains for agents.

### Submitter checklist
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/src/docker/docker-compose.yml
+++ b/src/docker/docker-compose.yml
@@ -25,7 +25,6 @@ services:
     image: jenkins/jenkins:lts
     ports:
       - "8080:8080"
-      - "50000:50000"
       - "50001:50001"
     volumes:
       - '/srv/docker/jenkins:/var/jenkins_home'
--- a/src/docker/docker-compose.yml
+++ b/src/docker/docker-compose.yml
@@ -26,8 +26,9 @@ services:
     ports:
       - "8080:8080"
       - "50000:50000"
+      - "50001:50001"
     volumes:
       - '/srv/docker/jenkins:/var/jenkins_home'
     environment:
-      - "JAVA_TOOL_OPTIONS=-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:50000"
+      - "JAVA_TOOL_OPTIONS=-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:50001"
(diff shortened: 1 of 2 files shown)


krisstern:
Hi @dodo920306 are you sure your proposed suggestion would work? We would like to debug port `50000` hence the original setting. Which debugger have you tried this on? And, why would you want to debug another port?

dodo920306:
> Hi @dodo920306 are you sure your proposed suggestion would work? We would like to debug port `50000` hence the original setting. Which debugger have you tried this on? And, why would you want to debug another port?

Accoring to https://www.jenkins.io/doc/book/security/services/#tcp-agent-listener-port
> This service is disabled by default in most packages, but the Docker images expose it at port 50000.

, using port 50000 as the JDWP debug port will make the docker container fail to start because of port conflict.

I use IntelliJ and it works pretty fine with the new port.

src/docker/docker-compose.yml
```diff
@@ -26,8 +26,9 @@ services:
     ports:
       - "8080:8080"
       - "50000:50000"
```
krisstern:
```suggestion
```

krisstern: It does not look like we need this anymore if we are to change ports away from the default TCP Agent Listener Port set at `50000`

krisstern (changes requested):

src/docker/docker-compose.yml
dodo920306: I think it should remain because someone may want to connect it as a Jenkins agent perhaps, but indeed, it's not necessary.

dodo920306: How do you think?

krisstern: They can add it to the config file themselves in that case, I don't think we need it here for the case considered.

krisstern (approved):
LGTM
Thanks for the PR @dodo920306!

PR #1877, opened 2026-03-29
Jackson3 migration

jonesbusy:
Jackson3 migration

### Testing done

mvn clean install

### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

PR #1881, opened 2026-03-31
Replace jersey3-api which still uses jackson 2 with jackson-jakarta-rs-json-provider

mmusenbr:
JacksonFeature was used from jersey3-api to provide the JackonJsonProvider. jersey3-api still uses jackson 2, which created a mismatch with the now used jackson3-api.

This fixes the behavior described/discussed in #1865, which was introduced by the switch to jackson3-api migration in #1877 
This pull request updates the project's dependency management and REST client configuration to use the Jackson JSON provider for Jakarta RS instead of Jersey's Jackson integration.

The plugin itself was migrated to jackson 3.x already. But with the JacksonFeature from jersey3-api, which still uses jackson 2.x, the Jackson 2 Json Provider is used in the RestEasy client. And therefore the default Jackson 2 Mapper is used, instead of the Jackon 3 ObjectMapper already provided by the plugin.

By replacing jersey3-api (Jackson 2.x) with jackson-jakarta-rs-json-provider the only Jackson version the plugin now uses through direct or transitive dependencies is Jackson 3.x now.

jackson-jakarta-rs-json-provider  should be replaced by a jenkins-api-plugin which provides a Jackson 3 JsonProvider, once available.

This will fix the current failing test in #1880.

### Submitter checklist
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed (in #1880)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -50,7 +50,7 @@
     <spotbugs.effort>Max</spotbugs.effort>
     <spotbugs.threshold>Low</spotbugs.threshold>
     <gitHubRepo>jenkinsci/${project.artifactId}</gitHubRepo>
-    <hpi.bundledArtifacts>asyncutil,jakarta.annotation-api,jakarta.validation-api,jboss-jaxrs-api_3.0_spec,jboss-logging,reactive-streams,resteasy-client,resteasy-client-api,resteasy-core,resteasy-core-spi</hpi.bundledArtifacts>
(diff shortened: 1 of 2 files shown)


krisstern:
Looks like we need to run 

```console
mvn spotless:apply
```

to fix the `Spotless` violations.

mmusenbr:
> Looks like we need to ... snip ... fix the `Spotless` violations.

omg, sorry for that. I was even manually moving that entry around to a more 'fitting' place. I was not aware that spotless orders the plugin section.
Thanks for pointing that out, Should be fixed now.

jonesbusy:
Thanks. I thought a bit more about it

I think it's more safe to stay on Jackson2 for now and wait for Jersey 4 with Jackson3 support

When https://github.com/eclipse-ee4j/jersey/issues/6036 is merged this probably deserve a Jersey 4 API plugin with dependency with Jackson3. Then this plugin can depends on Jersey 4 instead of Jersey 3

mmusenbr:
Sounds good. Definitely less intrusive, and no new 3rd party dependency.
Perfect, then we can close this pr.

